Issue #9, opened 2015-06-04
Problem when parse a part of html tree without root node


antonmi:
Example:
```html
<h2>Show user</h2>
<ul>
<li><strong>Name:</strong>Jim</li>
<li><strong>Age:</strong>25</li>
</ul>
<a href="/users">Back</a>
```
`Floki.parse/1` retruns only first node `{"h2", [], ["Show user"]}`
It is, of course, `:mochiweb_html.parse/1`. But can we make it better on Floki side?

philss:
@antonmi Thanks for pointing the issue.

Yes, it's a problem with the `:mochiweb_html.parse/1`. 
A possible way to solve it is to always include a "surrounding" tag before parse the HTML.
The idea is to inject the tag inside the `Floki.parse/1` when receiving the string.
Something like:

```elixir
def parse(html) do
  modified_html = "<floki_root_node>#{html}</floki_root_node>"
  {"floki_root_node", [], parsed_html} = :mochiweb_html.parse(modified_html)
  parsed_html
end
```

I made some manual tests and it's work. But more tests are needed.
I can fix this in the next week, or feel free to send a PR. Thanks!

antonmi:
I have used the same idea in my 'espec_phoenix' project!
The difference is that Floki.parse will return a list instead of a tuple.
But there is no effect to other functionality.

philss:
@antonmi Thanks for solving it! I'm closing now.
Just to register, the pull request that solved the problem was: https://github.com/philss/floki/pull/10

PR #10, opened 2015-06-04
#9 Problem when parse a part of html tree without root node

antonmi:


philss:
@antonmi Great! Thank you very much! :smiley:

philss:
@antonmi the version is already published. Update your mix.exs with the new version "0.2.1".

PR #11, opened 2015-06-05
Update .travis.yml

lowks:
Simplify further. Don't need `script: mix test`


philss:
@lowks Thanks! :smiley:

PR #19, opened 2015-09-03
Add hack solution for multiple selectors

### LINKED ISSUE #18: Cannot search with multiple selectors
With `@html` markup from `test/floki_test.exs`:

```
<html>
<head>
<title>Test</title>
</head>
<body>
  <div class="content">
    <a href="http://google.com" class="js-google js-cool">Google</a>
    <a href="http://elixir-lang.org" class="js-elixir js-cool">Elixir lang</a>
    <a href="http://java.com" class="js-java">Java</a>
  </div>
</body>
</html>
```

I am trying to search for an element with multiple classes (on the same element) with the selector `.js-cool.js-elixir` but Floki is not finding them. However, if I add a space between the classes, as if the second was a descendent of the first, `.js-cool .js-elixir`, it does return what I'm looking for but should not (as `js-elixir` is not descendent of `js-cool`).

Test demonstrating the error:
```
test "find elements with given multiple classes" do
  class_selector = ".js-cool.js-elixir"

  assert Floki.find(@html, class_selector) == [
    {"a", [
        {"href", "http://elixir-lang.org"},
        {"class", "js-elixir js-cool"}],
      ["Elixir lang"]}
  ]
end
```

licyeus:
This is a pretty gross solution for multiple selectors. I don't have a full understanding of the `Floki.Matchers` module, but I'm thinking that a full solution would involve creating a more robust representation of a selector, instead of just treating it as a string.

Closes #18

### DIFF
--- a/lib/floki/matchers.ex
+++ b/lib/floki/matchers.ex
@@ -57,9 +57,19 @@ defmodule Floki.Matchers do
   end
 
   def value_match?(attribute_value, selector_value) do
-    attribute_value
-    |> String.split
-    |> Enum.any?(fn(x) -> x == selector_value end)
+    selector_values = String.split(selector_value, ~r/[#.]/)
+
+    if length(selector_values) == 1 do
+      attribute_value
+      |> String.split
+      |> Enum.any?(fn(x) -> x == selector_value end)
+    else
+      attribute_value
+      |> String.split
+      |> Enum.all?(fn(x) ->
+        Enum.find(selector_values, fn(y) -> y == x end)
+      end)
+    end
   end
 
   def attribute_match?(attributes, attribute_name) do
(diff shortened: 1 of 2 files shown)


Lucasosf:
nice! :+1:

philss:
@licyeus thank you very much for the contribution! Please give me some time to try it. I think it could break the "ID selector" for some cases, but I'm not sure yet.

And you are right: we need a more robust representation of the selector (and the HTML tree as well), since today we can only search for "isolated" elements. Ideas on this are welcome.

I should check this weekend. Thanks again!

philss:
Hi @licyeus, thank you again for open the issue and the PR!

After thinking for a while, I decided to refactor `Floki` to accept a more robust combination of selectors like you mentioned.
In this refactor I also included the query that you implement in this PR. For this reason, I'm closing your PR. 

You can check the PR with those changes here: https://github.com/philss/floki/pull/22
Feel free to comment there.

PR #21, opened 2015-09-17
Update mix.exs to include :mochiweb in application configuration

EricDykstra:
Since `mochiweb` is a runtime dependency, it needs to be included in the application config when using exrm releases. Here's the error I was getting on my server if I didn't put `:mochiweb` in my own app's config.

```
** (exit) an exception was raised:
    ** (ErlangError) erlang error: :"module could not be loaded"
        :mochiweb_html.parse("<floki><!DOCTYPE html PUBLI….
```

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -12,7 +12,7 @@ defmodule Floki.Mixfile do
   end
 
   def application do
-    []
+    [applications: [:mochiweb]]
   end
 
   defp deps do


philss:
@EricDykstra great! I didn't know this is needed. Thanks! :smile: :+1:

EricDykstra:
Glad to help :blush:  Thanks for providing this great package!

Lucasosf:
:+1:

PR #34, opened 2015-10-09
Fix HTML comment in raw_html

glaucocustodio:
Node tree with comments was crashing before this fix.

### DIFF
--- a/lib/floki.ex
+++ b/lib/floki.ex
@@ -95,7 +95,7 @@ defmodule Floki do
   defp raw_html([], html), do: html
   defp raw_html(tuple, html) when is_tuple(tuple), do: raw_html([tuple], html)
   defp raw_html([value|tail], html) when is_binary(value), do: raw_html(tail, html <> value)
-  defp raw_html([{elem, value}|tail], html) when elem == :comment, do: raw_html(tail, html <> "<!--#{value}-->")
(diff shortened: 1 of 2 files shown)


lib/floki.ex
```diff
@@ -95,6 +95,7 @@ defmodule Floki do
   defp raw_html([], html), do: html
   defp raw_html(tuple, html) when is_tuple(tuple), do: raw_html([tuple], html)
   defp raw_html([value|tail], html) when is_binary(value), do: raw_html(tail, html <> value)
+  defp raw_html([{elem, value}|tail], html) when elem == :comment, do: raw_html(tail, html <> "<!--#{value}-->")
```
philss: @glaucocustodio here you don't need to use a guard. Instead, you can pattern match the `elem`:

```elixir
defp raw_html([{:comment, value}|tail], html), do: raw_html(tail, html <> "<!--#{value}-->")
```

glaucocustodio:
Oh man, sorry, I am not too used to pattern match. I've fixed.

philss:
@glaucocustodio no worries :+1:

philss:
@glaucocustodio Released in the version 0.6.1 :)

glaucocustodio:
:smile_cat:

PR #43, opened 2016-02-19
For DeepText <br> nodes are replaced by newline (\n)

maxneuvians:
Not sure if this is in line with the design philosophy for Floki, but I thought for a DeepText text extraction, \<br> nodes should be changed into \n. The reason being that it retains the layout intent of the text itself and a \<br> node is a terminal node in that it can't have children or text (compared to a \<p>).

Furthermore, the \n can easily be replaced in the resulting plain text if you want the old result, whereas you can't insert a newline at this point. 

Please feel free to reject the pull request if you don't agree with the change and I can maintain a fork of the library for my own purposes.

### DIFF
--- a/lib/floki/deep_text.ex
+++ b/lib/floki/deep_text.ex
@@ -29,6 +29,7 @@ defmodule Floki.DeepText do
     end
   end
   defp get_text({:comment, _}, acc), do: acc
+  defp get_text({"br", _, _}, acc), do: acc <> "\n"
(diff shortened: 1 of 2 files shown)


philss:
@maxneuvians Thanks for the PR! I agree with you. :+1:

philss:
@maxneuvians This fix is available in the new version 0.7.2. Thanks again!

maxneuvians:
No problem - my pleasure!

PR #47, opened 2016-05-19
Support rendering boolean attributes

iamvery:
## Problem

The function `Floki.raw_html/1` does not handle boolean attributes.

A boolean attribute is one who's lack of corresponding value indicates truthiness. For example:
```html
<div hidden></div>
```

Since attributes in Floki's AST are represented as a 2 element tuple, there is no clear way to render an AST with boolean values.

## Solution

Update the rendering code to support rendering attributes that are not a 2 element tuple.

```elixir
Floki.raw_html({"div", ["hidden"], []})
"<div hidden></div>"
```

Thankfully this does not change any of the existing behavior in Floki, it just extends what is there to support another form of attribute when rendering an AST.

## A Note About Parsing

The above described AST, `{"div", ["hidden"], []}`, is not a valid AST produced by mochiweb_html (which is the parser used internally by Floki). In fact, mochiweb_html:parse/1 would produce something like:

```elixir
:mochiweb_html.parse("<div hidden></div>")
{"div", ["hidden", "hidden"], []}
```

After thinking about this (and trying and failing to "fix" mochiweb_html), I decided it might be OK to only produce a change in Floki (if only "for now"). This is based on the fact that mochiweb_html is an implementation detail. Producing HTML _from_ the AST is a different responsibility, which is in fact owned by the Floki lib.

What do you think?

### DIFF
--- a/lib/floki.ex
+++ b/lib/floki.ex
@@ -106,6 +106,7 @@ defmodule Floki do
   end
 
   defp build_attrs({attr, value}, attrs), do: ~s(#{attrs} #{attr}="#{value}")
+  defp build_attrs(attr, attrs), do: "#{attrs} #{attr}"
 
   defp tag_for(type, attrs, _children) when type in @self_closing_tags do
     case attrs do
--- a/lib/floki.ex
+++ b/lib/floki.ex
@@ -101,9 +101,11 @@ defmodule Floki do
 
   defp tag_attrs(attr_list) do
     attr_list
-    |> Enum.reduce("", fn({attr, value}, attrs) -> ~s(#{attrs} #{attr}="#{value}") end)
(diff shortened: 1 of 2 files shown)


philss:
@iamvery I agree with you. In the (near) future I want to not depend on mochiweb and be able to control this kind of thing (the parsing side).

Thanks a lot for the PR! 😃

iamvery:
Thanks @philss! Glad you found this useful 😄 

❤️

philss:
@iamvery This change is available in the version [0.8.1](https://github.com/philss/floki/blob/master/CHANGELOG.md#081---2016-05-20). Please give a try 😃

iamvery:
Perfect! Thank you so much for cutting a new release so quickly. Wow!

PR #51, opened 2016-05-24
add optional separator to text function

rochdi:
@philss

### DIFF
--- a/lib/floki/flat_text.ex
+++ b/lib/floki/flat_text.ex
@@ -6,7 +6,7 @@ defmodule Floki.FlatText do
 
   @type html_tree :: tuple | list
 
-  @spec get(html_tree) :: binary
+  @spec get(html_tree, binary) :: binary
 
   @doc """
   Get text nodes from first level of HTML nodes.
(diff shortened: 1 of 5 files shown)


test/floki/deep_text_test.exs
```diff
@@ -9,6 +9,12 @@ defmodule Floki.DeepTextTest do
     assert Floki.DeepText.get(node) == "Google"
   end
 
+  test "text from simple node with separator" do
+    node = {"a", [], ["Google"]}
+
+    assert Floki.DeepText.get(node, " ") == " Google"
```
philss: I'm not sure if it's useful to add a space in the beginning of the result. Seems unnecessary.

I believe the separator should work like `Enum.join/2`.
E.g.:

```elixir
Enum.join(["Hello", "World"], " ")
=> "Hello World"
```

@rochdi What do you think?

mix.lock
```diff
@@ -1,6 +1,6 @@
-%{"earmark": {:hex, :earmark, "0.2.1", "ba6d26ceb16106d069b289df66751734802777a3cbb6787026dd800ffeb850f3", [:mix], []},
-  "ex_doc": {:hex, :ex_doc, "0.11.5", "0dc51cb84f8312162a2313d6c71573a9afa332333d8a332bb12540861b9834db", [:mix], [{:earmark, "~> 0.1.17 or ~> 0.2", [hex: :earmark, optional: true]}]},
-  "inch_ex": {:hex, :inch_ex, "0.5.1", "c1c18966c935944cbb2d273796b36e44fab3c54fd59f906ff026a686205b4e14", [:mix], [{:poison, "~> 1.5 or ~> 2.0", [hex: :poison, optional: false]}]},
+%{"earmark": {:hex, :earmark, "0.2.1"},
```
philss: can you update your local `hex`? or please revert this file.

lib/floki/flat_text.ex
```diff
@@ -6,33 +6,40 @@ defmodule Floki.FlatText do
...
+  @spec get(html_tree, binary) :: binary
 
   @doc """
   Get text nodes from first level of HTML nodes.
 
 
   ## Examples
 
-      iex> Floki.FlatText.get([{"a", [], ["The meaning of life is...", {"strong", [], ["something else"]}] }])
+      iex> Floki.FlatText.get([{"a", [], ["The meaning of life is...", {"strong", [], ["something else"]}] }], "")
```
philss: This example can remain the previous version, without the empty string (since it's already being showed in the line below).

rochdi: Ok

mix.lock
rochdi: Done

test/floki/deep_text_test.exs
rochdi: Agree, need to find an elegant way to achieve this

rochdi:
@philss 
* added a text_util module to avoir having a space at the beginning of the result
* update based on your comments the test in the doc

philss:
Hi @rochdi! Sorry for the delay. The functionality is OK for me. One thing I'm not pretty sure about is that `TextUtil` module. This is because the problem could have been solved using a simpler approach, like pattern matching.

```elixir
# Instead of:
defp get_text(text, acc, sep) when is_binary(text), do: TextUtil.join_text(@intial_acc, acc, text, sep)

# We could:
defp get_text(text, "", sep) when is_binary(text), do: text
defp get_text(text, acc, sep) when is_binary(text), do: acc <> text <> sep
```

Of course this would add a little duplication, but seems more easy to understand.

I can merge and fix the way I think, or I can wait another version from you. What do you prefer?

And thank you for this! 😄

rochdi:
@philss I removed TextUtil module, let keep it clean.
thanks

philss:
@rochdi I will release a new version with this tomorrow. Thank you! 😃

darrenburns:
Hi @philss, when will you be releasing a version with this feature? It's very useful :)

philss:
@darrenburns @rochdi sorry for the delay guys, it's new alive as v0.9.0. Thanks!

PR #54, opened 2016-07-05
Update README to use the latest floki version

alexwbuschle:


philss:
@alexwbuschle Thanks! 😃

PR #55, opened 2016-07-30
Re-word: array -> list

gmile:
Noticed the mention of `array` in the docs. I think it makes sense to reword this everywhere.


philss:
@gmile Thank you! 😃

PR #57, opened 2016-07-31
[WIP] Fix immediate children selector

gmile:


lib/floki/finder.ex
```diff
@@ -95,7 +95,12 @@ defmodule Floki.Finder do
           nil -> [n|res_acc]
           _ ->
             {_, _, children_nodes} = n
-            traverse(children_nodes, sibling_nodes, selector.combinator.selector, res_acc)
+
+            if selector.combinator.match_type == :child do
```
gmile: We much check if the next element is worth "diving into". If it is – continue the "child selector combo" by calling just `traverse_child(children_nodes...)`.

If the combinator is a different selector – this means the combo is successfully over (e.g. entire chil chain was resolved), and we can go back to normal traversal via `traverse(nodes...)`.

gmile:
I _think_ I'm close to fix this one. But after I added a more complex test case, the test is now failing again (duplicating the expected result). Looking into it right now, but any help is more than appreciated!

lib/floki/finder.ex
```diff
@@ -95,7 +95,12 @@ defmodule Floki.Finder do
...
+              traverse_child(children_nodes, sibling_nodes, selector.combinator.selector, res_acc)
+            else
+              traverse(nodes, sibling_nodes, selector, res_acc)
```
philss: @gmile I think the problem is that you are passing the `selector`, and not `selector.combinator.selector` here. 
Except for that, this seems to be correct.

gmile: @philss that didn't help. I think I'm doing it wrong when passing `nodes`, as `nodes` is something Enum.reduce is still reducing on

philss: @gmile I'm still thinking about this. Not sure what is going wrong. One suggestion is to isolate that `case` statement from the line 70, and try to reuse that here.

I suspect that something is wrong with the way `sibling_nodes` are traversed (this explain why the `div > p > span > img + img` is not working).

I'm gonna check this tomorrow night again. Thanks!

gmile: I extracted traversal selection https://github.com/philss/floki/pull/59, although it hadn't have a direct impact on what's going on here.

gmile:
Will be cleaned up when (if) #60 is merged.

philss:
@gmile I'm gonna release a new version with the fix from #60.

Also, please update your branch with master. Thanks!

gmile:
@philss done. This PR now only contains a single failing test, that was discovered while working on the previous now-fixed issue.

philss:
@gmile I'm closing this one in order to be able to open another PR.

Issue #46, opened 2016-03-04
How to find tag with namespace?


fan-y-ouchi:
```
iex(1)> "<ns:tag>xxx</ns:tag>" |> Floki.find("ns:tag")
Unknown token ':'. Ignoring.
[]
```

The jQuery can do as follows:
`$("ns\\:tag")`

philss:
Hi @fan-y-ouchi! This feature was not present yet. But is now available in the version `0.8.0` of Floki. :smiley: 

Please update and check again (the same syntax of your example).

fan-y-ouchi:
Thank you! :smile:

gmile:
@philss will this not conflict with possible future implementation of `:first-child`, `:nth-child()` etc.? Per ** 6.1.1. Type selectors and namespaces** from [here](https://www.w3.org/TR/css3-selectors/#typenmsp), one must target namespaced tags by means of dash symbol `|`.

@philss what do you think?

philss:
@gmile wow! I didn't know that. I think I misunderstood this namespace selector. Thank you for pointing out. This need to be fixed. I'm gonna reopen this.

philss:
Closed again by #58.

philss:
@fan-y-ouchi @gmile the new version was released in v0.10.0. Thanks!

PR #58, opened 2016-07-31
Fix search of elements with namespace

philss:
It fixes a wrong selector when searcing elements using namespace, like
is described in https://www.w3.org/TR/css3-selectors/#univnmsp.

Now you can search for elements inside namespace with:

    Floki.find(html, "ns | tag")

This closes the issue #46 again.


philss:
@gmile This PR fixes the issue that you commented in #46. Feel free to comment if you find any issues :)

gmile:
@philss looks great, thank so much!

philss:
@gmile thank you! 😃

PR #59, opened 2016-08-01
Extract traversal selection

gmile:


lib/floki/finder.ex
```diff
@@ -68,18 +68,7 @@ defmodule Floki.Finder do
...
-              :child ->
-                traverse_child(children_nodes, sibling_nodes, combinator.selector, acc)
-              :sibling ->
-                traverse_sibling(children_nodes, sibling_nodes, combinator.selector, acc)
-              :general_sibling ->
-                traverse_general_sibling(children_nodes, sibling_nodes, combinator.selector, acc)
-              other ->
-                raise "Combinator of type \"#{other}\" not implemented"
-            end
+            traverse_using(combinator.match_type, children_nodes, sibling_nodes, combinator.selector, acc)
```
philss: @gmile I think we can remove one argument to this function by passing the `combinator` and accessing `match_type` and `selector` from inside the function. What do you think?

gmile: @philss sure, I just pushed the change

philss:
@gmile thank you! :shipit:

PR #60, opened 2016-08-02
Ensure immediate children selector is working

gmile:
The test was extracted from #57. That PR still has a test that fails, but that's a different issue. This PR at least addresses the original problem with immediate children.


philss:
@gmile Thanks! :D

philss:
@gmile this was released in version v0.10. Thanks again! 😃

PR #62, opened 2016-08-20
Fix sibling after immediate children

philss:
It fixes the search for sibling after a chain of immediate children.
The problem was in the siblings that were been passed to the traverse
child function. We had to fetch the siblings again after matching a
node.

Related to PR #57


philss:
@gmile Please check this.

Issue #65, opened 2016-09-02
Supporting transforms for Floki nodes!


bobjflong:
Hi! Thanks so much for Floki!

I have a need to transform Floki nodes. For example, to turn all relative links into absolute links:

```elixir
html = "<p><a href='/foo'>click!</a></p>"
Floki.transform(html, "a", fn({"a", [href: x], xs}) -> {"a", URI.merge("https://base.com", x), xs} end)
# => [{"p", [], [{"a", [{"href", "https://base.com/foo"}], ["click!"]}]}]
```

Instead of requiring the implementation of a recursive search, do you think there is space in the Floki API for a `transform` function like above? I'd be up for cutting a PR myself. Just wondering if you have any thoughts.

philss:
Hi @bobjflong! Thank you for the PR and the issue! 😃  I couldn't check this with attention yet, but I will check soon. :)

PR #66, opened 2016-09-02
first attempt at transform function

bobjflong:
fixes https://github.com/philss/floki/issues/65

Just uses Elixir's pattern matching on the html tree structure. This is how I'd implement the function outside of the library. Maybe there's a more efficient way using Floki's selector engine?

### DIFF
--- a/lib/floki.ex
+++ b/lib/floki.ex
@@ -149,6 +149,16 @@ defmodule Floki do
     Finder.find(html_tree, selector)
   end
 
+  def transform(html_tree_list, transformation) when is_list(html_tree_list) do
+    html_tree_list |> Enum.map(fn
+      html_tree ->
+        Finder.find_and_transform(html_tree, transformation)
+      end)
+  end
+  def transform(html_tree, transformation) do
+    Finder.find_and_transform(html_tree, transformation)
+  end
+
   @doc """
   Returns the text nodes from a HTML tree.
   By default, it will perform a deep search through the HTML tree.
--- a/lib/floki.ex
+++ b/lib/floki.ex
@@ -152,11 +152,11 @@ defmodule Floki do
   def transform(html_tree_list, transformation) when is_list(html_tree_list) do
     html_tree_list |> Enum.map(fn
       html_tree ->
-        Finder.find_and_transform(html_tree, transformation)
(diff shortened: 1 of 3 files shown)


rstacruz:
Aw yiss, this is very useful.

rstacruz:
I'd also love a way to pass a selector to the transformation, just like `Floki.find`:

```ex
Floki.transform(html, "a[href^=\"http\"]", &add_class/1)

def add_class({tag, attrs, children}) do
  {tag, [class: "external"] ++ attrs, children}
end
```

philss:
@bobjflong Thank you for the PR! And sorry for the delay!

I'm gonna merge this, and try to improve a little bit.

The ideal solution is to make it work by updating the HTML tree. Not sure if it's possible in the current code.

philss:
@rstacruz Cool! This is the ideal API.

But I'm not sure if it's clear that we are transforming the HTML tree or only the results of selector.

```html
<html>
  <body>
    <h1>Example</h1>
    <ul>
      <li><a href="http://external.com">Link to external</a></li>
      <li><a href="/home">Link to home</a></li>
    </ul>
  </body>
</html>
```

The Elixir code:

```elixir
Floki.transform(html, "a[href^=\"http\"]", &add_class/1)

def add_class({tag, attrs, children}) do
  {tag, [class: "external"] ++ attrs, children}
end
```

1. Update the tree will return:

```elixir
{"html", [],
 [{"body", [],
   [{"h1", [], ["Example"]},
    {"ul", [],
     [{"li", [],
       [{"a", [{:class, "external"}, {"href", "http://external.com"}], ["Link to external"]}]},
      {"li", [], [{"a", [{"href", "/home"}], ["Link to home"]}]}]}]}]}
```

2. Update and return only results:

```elixir
[{"a", [{:class, "external"}, {"href", "http://external.com"}], ["Link to external"]}]
```

In my opinion, the user is expecting the first option. What do you think?

philss:
@bobjflong This is released in version [v0.11.0](https://hex.pm/packages/floki/0.11.0). Thanks again!

PS: I'm working in a version to receive the selector update and the tree, as we talked here. I hope to finish this soon. I will let you know. cc @rstacruz

Issue #68, opened 2016-10-12
mochiweb_html breaks building a release when other dependancies include mochiweb


aphillipo:
Hi @philss - thanks so much for Floki it's awesome...

I'm in a bit of a pickle with mochiweb_html conflicting with deps that use the mochiweb package.

I understand your reasoning only including the mochiweb_html parser rather than all of the other stuff with Floki, however when trying to build a release it causes the following error:

==> Failed to build release:
    
    Duplicated modules: 
    	mochiweb_html specified in mochiweb and mochiweb_html
    	mochiweb_charref specified in mochiweb and mochiweb_html
    	mochiutf8 specified in mochiweb and mochiweb_html
    	mochinum specified in mochiweb and mochiweb_html

Given that releases are borked and I'd have to fork your library (as would anyone building a release that includes mochiweb in the deps) would you consider this optimisation less desirable and go over to using full mochiweb?

Thanks again!!!

philss:
Hi @aphillipo! Thank you!! I didn't know about this problem, and I totally agree with using the full `mochiweb` package now.

PS: I'm the publisher of that `mochiweb_html` package. Now that you found the issue, I think it makes sense to update [that package](https://github.com/philss/mochiweb_html) to use another namespace.

PR #69, opened 2016-10-12
changed to full mochiweb package to stop duplicates inside a release

aphillipo:
You can see the reasoning inside #68 

Thanks!

EDIT: I have also confirmed that this works and that the release builds and works...


philss:
@aphillipo This is released in version [v0.11.0](https://hex.pm/packages/floki/0.11.0). Thanks again!

Issue #84, opened 2017-01-18
Floki.find returning unmatching nodes


mmmries:
I tried to update a project to the new `0.12.0` release, but ran into a bug. Here is a minimal reproduction of the error:

```elixir
    html = """
      <html>
      <body>
      <div id="messageBox" class="legacyErrors"><div class="messageBox error"><h2 class="accessAid">Error Message</h2><p>There has been an error in your account.</p></div></div>
      <div id="main" class="legacyErrors"><p>Separate Error Message</p></div>
      </body>
      </html>
      """
    html |> Floki.parse |> Floki.find(".messageBox p") |> Floki.text
    # => "There has been an error in your account.Separate Error Message"
```

The `.messageBox p` selector should only return the first `p` node, but it actually returns both of the `p` nodes and the `Floki.text` call ends up returning the text from both of those nodes.

mmmries:
BTW thanks for doing this project. It has been very useful so far and I'd be happy to spend some time trying to fix this bug if you can point me in the right general direction of where I should look.

philss:
Hi @mmmries! 

Thanks for the report. This is definitely a bug. Not sure why, but is selecting a wrong node.
I'm investigating this.

philss:
@mmmries I found the problem, but I'm not sure how to fix yet. It is here: https://github.com/philss/floki/blob/master/lib/floki/finder.ex#L202-L210

We should take all the children ids based on the list of ids that are between the current node and the sibling node. This works when there is a sibling, but don't work when the node is the last node inside his parent.

I have one idea that I will try to implement. Feel free to give a try too.

Thanks!

mmmries:
@philss thanks for pointing me in the right direction. I'll take a look and see if I can better understand that section of code specifically. Hopefully I can make an attempt at a fix before you get there so you can take me to 🏫

philss:
@mmmries nice! One way to make sure it takes all descendant nodes is to traverse the nodes that are below the current node and adding them to the stack right in that process.
This would avoid the idea of taking the ids of what we assume it's a subtree, which theoretically is simpler. But would be a little bit harder to implement since we would have to traverse and accumulate the results.

philss:
@mmmries This fix was released in v0.12.1. Thank you again!

mmmries:
Thanks for the release and pointing me in the right direction to work on a fix :blue_heart: :yellow_heart: :green_heart:

PR #85, opened 2017-01-20
get descendant node ids recursively rather than using a range

mmmries:
Attempt at fixing #84 

Rather than finding all node_ids in a given range (using the next sibling as an upper-bound), find all the children by recursively walking down the tree looking for leaf nodes.

I've also added the test case from #84, although I'm not sure that putting that test in `test/floki_test.exs` is the best place for it.

/cc @philss

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,10 @@ matrix:
       elixir: 1.3.4
     - otp_release: 19.0
       elixir: 1.3.4
+    - otp_release: 18.3
+      elixir: 1.4.0
+    - otp_release: 19.0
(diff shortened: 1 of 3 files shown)


mmmries:
Not sure if it was appropriate to add to this PR, but I went ahead and added elixir `1.4.0` under OTP 18.3 and OTP 19.0 just to make sure we are keeping up support for that.

philss:
@mmmries wow! That was fast! Thanks! 😃 

I have to release two versions with this fix. I will do that probably today. I will ping you back. Thanks again!

PR #86, opened 2017-01-22
Duplicate reversed descendant matches

mmmries:
Thanks for all of your help with #85 

I tried upgrading to `v0.12.1` in my project and ran into two small issues. This PR attempts to fix one of those two.  `Finder` was returning some duplicate nodes and the nodes were not in the same order that they occured in the html tree.

I've added a spec to highlight this and tried to address it in `Finder`.


philss:
@mmmries The problem with duplicated nodes was present before. Thank you for fixing both.

philss:
@mmmries released under `0.13.1` :)

PR #87, opened 2017-01-22
Text reversing

mmmries:
I first noticed this problem manifest when I would do:

```
iex> Floki.text([{"p", [], ["1"]},{"p", [], ["2"]}])
"21"
```

The order of the text was reverse of the order of the nodes being passed in. Digging in further it turned that that the reversing was happening during the `FilterOut` step, and digging in past that it turned out be caused when `Finder` constructs an `HTMLTree` struct because it was recording the `root_node_ids` in reverse order of what got passed in.

I've added a few unit tests as I was tracking this down, but we can remove most of them if you prefer. It might only be useful to keep the unit test added to `HTMLTree` since that was the root cause in the end, but I figured you might want to keep the coverage around from the other tests just in case.

### DIFF
--- a/lib/floki.ex
+++ b/lib/floki.ex
@@ -190,6 +190,9 @@ defmodule Floki do
       iex> Floki.text([{"div", [], ["hello world"]}])
       "hello world"
 
+      iex> Floki.text([{"p", [], ["1"]},{"p", [], ["2"]}])
+      "12"
+
   """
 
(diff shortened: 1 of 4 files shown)


philss:
@mmmries great! Thank you! 😃

philss:
@mmmries released under `0.13.1` too.

PR #88, opened 2017-02-02
Only Mochiweb_html

mendrugory:
I am working in a project where I am using Floki and rabbit_common libraries. Both libraries depend on mochijson2 and it caused dependency issues. After checking Floki, I saw that it only uses mochiweb_html, therefore I have tried to run Floki with only this dependency. Everything is working fine for me, hence; I propose to make lighter the load of 3rd party libraries in Floki.

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -9,7 +9,7 @@ defmodule Floki.Mixfile do
      name: "Floki",
      version: @version,
      description: @description,
-     elixir: ">= 1.2.0",
+     elixir: ">= 1.4.0",
      package: package(),
      deps: deps(),
      source_url: "https://github.com/philss/floki",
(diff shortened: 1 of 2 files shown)


philss:
Hi @mendrugory. Thank you for the PR!

We had the same problem in #68, but with the inverse cause: one was using `mochiweb` and Floki was depending on `mochiweb_html` (which I published on Hex).
I like the idea, but I think we should avoid those modules conflicts. One alternative is to rename modules in https://github.com/philss/mochiweb_html. Another alternative, is rename that fork with a name that can't conflict, and add the name as sufix of modules inside.

What do you think?

mendrugory:
Hi @philss, 

It is true that we will always have problems with some dependencies or others. Both could be good, but I would choose that one that would have less maintenance ;)

Issue #96, opened 2017-02-27
:not selector not working with parents


Eiji7:
see: #95

philss:
@Eiji7 good catch! I'm investigating. Thank you!

philss:
@Eiji7 The fix was released in version 0.15.0. Thank you! :)

PR #97, opened 2017-02-28
Fix search for selectors with combinators and pseudo-classes

philss:
It fixes issue #96.
And also fixes problems with other combinators.


philss:
Hi @Eiji7. Thanks for finding the bug.
This PR fixes it.
I gonna merge it right after AWS S3 come back from outage and run the build ;)

Eiji7:
@philss: look like ok now

Issue #100, opened 2017-03-18
:not selector cannot handle pseudo-classes with arguments


mischov:
**Problem**:
Definition `PARENTESIS = \([^)]*\)` from the selector lexer captures the next closing paren but not the *balanced* closing paren, which mis-tokenizes selectors such as `:not(:nth-child(2))`

This causes errors like
```elixir
> html = """
  <ol>
    <li><a>1</a></li>
    <li><a>2</a></li>
    <li><a>3</a></li>
  </ol>
  """
"..."

> Floki.find(html, "li:not(:nth-child(2)) a")
[{"a", [], ["1"]}, {"a", [], ["2"]}, {"a", [], ["3"]}] # Should only return 1st and 3rd
```

**Suggestion**: 
(I think) Regex lacks the state to capture balanced parens, so the tokenizer should just return closing parens and the parser should make the appropriate decisions.

Something like: https://github.com/mischov/meeseeks/blob/6863e9dd79a42b2ef0a3415d976e6d913b311069/lib/meeseeks/selector/parser.ex#L220-L240

philss:
@mischov thank you for the report! 😃 
I think your suggestion may work. I will try to fix this soon. Thanks!

philss:
@mischov the fix was released in version 0.16.0 as well. Thank you!

Issue #92, opened 2017-02-05
Add support for `Floki` and / or jQuery special selectors.


Eiji7:
Add `Floki` and / or jQuery special selectors. More info at [Category: Selectors at jQuery docs](https://api.jquery.com/category/selectors/).
Examples:
```css
:floki-has-text("     a text       ")
:floki-has-trimmed-text("a text")
:jquery-empty
```
Add more `Floki` special selectors suggestions as a comments for this issue.
**Note**: I split my comment and add its half as current issue on @philss request at #82 issue.

philss:
I will close this one and open individual issues for each selector we want. Thank you.

PR #101, opened 2017-03-31
add support for contains selector

buhman:
This adds support for the `:contains` pseudo-class selector from jQuery.

Usage is demonstrated in the `contains pseudo-class` test.

### DIFF
--- a/lib/floki/selector/pseudo_class.ex
+++ b/lib/floki/selector/pseudo_class.ex
@@ -6,7 +6,7 @@ defmodule Floki.Selector.PseudoClass do
   # Represents a pseudo-class selector
   defstruct name: "", value: nil
 
-  alias Floki.HTMLTree.HTMLNode
+  alias Floki.HTMLTree.{HTMLNode, Text}
 
   def match_nth_child?(_, %HTMLNode{parent_node_id: nil}, _), do: false
   def match_nth_child?(tree, html_node, %__MODULE__{value: position}) when is_integer(position) do
@@ -23,6 +23,17 @@ defmodule Floki.Selector.PseudoClass do
   def match_nth_child?(_, _, %__MODULE__{value: expression}) do
     Logger.warn("Pseudo-class nth-child with expressions like #{inspect expression} are not supported yet. Ignoring.")
     false
+  end
+
+  def match_contains?(tree, html_node, %__MODULE__{value: value}) do
+    res = Enum.find(html_node.children_nodes_ids, fn(id) ->
+      case Map.get(tree.nodes, id) do
+        %Text{content: content} -> content =~ value
+        _ -> false
+      end
+    end)
+
+    res != nil
   end
 
   defp node_position(tree, html_node) do
--- a/src/floki_selector_lexer.xrl
+++ b/src/floki_selector_lexer.xrl
@@ -24,6 +24,7 @@ Rules.
 \({ODD}\)                            : {token, {pseudo_class_odd, TokenLine}}.
 \({EVEN}\)                           : {token, {pseudo_class_even, TokenLine}}.
 \({PSEUDO_PATT}\)                    : {token, {pseudo_class_pattern, TokenLine, remove_wrapper(TokenChars)}}.
+\({QUOTED}\)                         : {token, {pseudo_class_quoted, TokenLine, remove_wrapper(remove_wrapper(TokenChars))}}.
 {W}*\)                               : {token, {close_parentesis, TokenLine}}.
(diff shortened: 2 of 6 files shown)


buhman:
> `jq-contains`

Well, this one isn't *exactly* the same as jquery's `contains`--it won't do `contains(unquoted)`, which jquery supports. `floki-contains` seems a bit long; what about something shorter like `fl-contains`?

philss:
@buhman thank you again! Tomorrow I gonna check this PR.

Eiji7:
@buhman: your first comment:
> This adds support for the :contains pseudo-class selector from jQuery.

then
> Well, this one isn't exactly the same as jquery's contains--it won't do contains(unquoted), which jquery supports, so that might be a bit misleading.

It's why I suggested `jq-contains` firstly ...

> floki-contains seems a bit long; what about something shorter like fl-contains?

I think that @philss should decide about it. I just suggest to add a prefix for non-standard selectors like I suggested in #92.

philss:
@buhman I agree with @Eiji7 that we need a prefix. Even if the behavior was the same as jQuery, I think it's better to add a prefix that identifies `floki`. I like `fl-contains` 👍 .

Please update your branch with master. There was a move of the `pseudo-class` matcher in this PR: https://github.com/philss/floki/pull/104

lib/floki/selector/pseudo_class.ex
```diff
@@ -25,6 +25,17 @@ defmodule Floki.Selector.PseudoClass do
     false
   end
 
+  def match_contains?(tree, html_node, %__MODULE__{value: value}) do
+    res = Enum.filter(html_node.children_nodes_ids, fn(id) ->
```
philss: You could use `find` instead of `filter`, since we need to know if one or more `Text` nodes matches.

Another thing: it's more correct to check if the `value` is inside the `content` of the `Text` nodes.
The way it is, it will only match nodes that have only one word.

Example:

```elixir
html = """
<div>
  <p>roses are red, violets are blue</p>
</div>
"""

Floki.find(html, ":fl-contains(red)")
```

In this example, it won't match the "<p>" because the text node will include all the text (not only "red").

A way to match this would be something like:

```elixir
case Map.get(tree.nodes, id) do
  %Text{content: content} -> String.split(content, " ") |> Enum.any?(& &1 == value)
  _ -> false
end
```

Of course we have some ed-cases like the `,` (comma), and punctuation. We have to think in a way to avoid that.

src/floki_selector_lexer.xrl
```diff
@@ -22,6 +22,7 @@ Rules.
 \({ODD}\)                            : {token, {pseudo_class_odd, TokenLine}}.
 \({EVEN}\)                           : {token, {pseudo_class_even, TokenLine}}.
 \({PSEUDO_PATT}\)                    : {token, {pseudo_class_pattern, TokenLine, remove_wrapper(TokenChars)}}.
+\({QUOTED}\)                         : {token, {pseudo_class_quoted, TokenLine, remove_wrapper(remove_wrapper(TokenChars))}}.
```
philss: I had to change the parsing of pseudo-class values because of this issue #100. I'm not sure but seems that this line does the same it was doing before I fix in #104 . Can you confirm this?

buhman: I don't think so.

It seemed cleaner to me to disambiguate nested css expressions from literals/strings, so I made `:psuedo_class_quoted`, and also nicer-looking inside `selector_parser`.

lib/floki/selector/pseudo_class.ex
buhman: I added a test for this as well. 

> String.split(content, " ") |> Enum.any?(& &1 == value)

I think substring match (`=~`) might be a little better, so you can do things like `fl-contains('violets are'),

src/floki_selector_lexer.xrl
mischov: Because {QUOTED} will capture from the first quotation mark to the next, something like `("searching for this (and that)")` should be fine (though I didn't test to make sure).

philss: You are right guys. It works good. 👍 

@buhman I think there is no way to use groups in leex today. What would make a little bit fancier is to write another function that removes double wrappers. But I think it's fine this way.

philss:
@buhman great work! 👏  Thank you! 😃

philss:
@mischov @Eiji7 also thank you both for the review! 😃

philss:
@buhman @Eiji7 @mischov this was released in version 0.16.0.

PR #103, opened 2017-03-31
finder: add support for selectors that only include a pseudo-class

### LINKED ISSUE #102: psuedo-class selectors require non-psuedo-class selectors
I noticed [here](https://github.com/philss/floki/blob/master/lib/floki/finder.ex#L93) that selectors that only specify a psuedo-class will never match. For example:

```css
:first-child
```

Should match the first child of every element.

I was thinking the least-refactoring-required way to do this would be something like:

```elixir
defp selector_match?(tree, html_node, selector = %Selector{id: nil, type: nil, classes: [], attributes: [], namespace: nil}) do
  pseudo_class_match?(tree, html_node, selector)
end
```

Does this seem sane?

buhman:
This also tweaks node_position logic so that it can be called on non-HTMLNode elements.

Fixes #102.

### DIFF
--- a/lib/floki/selector/pseudo_class.ex
+++ b/lib/floki/selector/pseudo_class.ex
@@ -32,9 +32,10 @@ defmodule Floki.Selector.PseudoClass do
                          |> filter_only_html_nodes(tree.nodes)
                          |> Enum.with_index(1)
 
-    {_node_id, position} = Enum.find(children_nodes_ids, fn({id, _}) -> id == html_node.node_id end)
-
-    position
+    case Enum.find(children_nodes_ids, fn({id, _}) -> id == html_node.node_id end) do
+      {_node_id, position} -> position
+      _ -> nil
+    end
   end
 
   defp filter_only_html_nodes(ids, nodes) do
--- a/test/floki_test.exs
+++ b/test/floki_test.exs
@@ -540,6 +540,16 @@ defmodule FlokiTest do
     assert first_result == second_result
   end
 
+  test "pseudo-class selector only" do
+    expected = [
+      {"channel", [], [{"title", [], ["A podcast"]}, {"link", [], ["www.foo.bar.com"]}]},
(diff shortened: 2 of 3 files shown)


philss:
Hi @buhman! Thank you for your PR. I'm gonna investigate it, since it's very similar to what I did in this PR that I just opened: https://github.com/philss/floki/pull/104.

test/floki_test.exs
```diff
@@ -540,6 +540,17 @@ defmodule FlokiTest do
...
   end
 
+  test "pseudo-class selector only" do
+    expected = [
+      {"channel", [], [{"title", [], ["A podcast"]}, {"link", [], ["www.foo.bar.com"]}]},
+      {"title", [], ["A podcast"]},
+      {"title", [], ["Another podcast"]}
+    ]
+
+    IO.inspect Floki.find(@xml, ":first-child")
```
buhman: woops

lib/floki/selector/pseudo_class.ex
```diff
@@ -32,9 +32,10 @@ defmodule Floki.Selector.PseudoClass do
                          |> filter_only_html_nodes(tree.nodes)
                          |> Enum.with_index(1)
 
-    {_node_id, position} = Enum.find(children_nodes_ids, fn({id, _}) -> id == html_node.node_id end)
-
-    position
+    case Enum.find(children_nodes_ids, fn({id, _}) -> id == html_node.node_id end) do
```
philss: We don't need this case because will have a match case for nodes that don't have parents (this is the only case `Enum.find` would not find).
 
I will merge with this and fix in my branch.

philss:
@buhman @Eiji7 Thank you!
Merging this one. I will remove the duplications in my branch/PR.

philss:
@buhman @Eiji7 this was released in version 0.16.0.

PR #105, opened 2017-04-06
update pseudo not selector to allow multiple selections in ()

jjcarstens:
While trying to utilize the new pseudo `not` selector, I was trying to do a selection with multiple _or_ selectors in the `not` pseudo class. i.e. `a.link:not(.bar, [style=baz])`

However, `finder.ex` was splitting the string at `,` which broke up all subsequent token parsing. So I went ahead and mocked this up to support this use-case.

Happy to change what is needed to meet any style preference or design that's not currently met. Let me know what you think.

@philss @mmmries

### DIFF
--- a/lib/floki/selector.ex
+++ b/lib/floki/selector.ex
@@ -23,7 +23,7 @@ defmodule Floki.Selector do
                               classes: [],
                               attributes: [],
                               namespace: nil,
-                              pseudo_class: nil,
+                              pseudo_classes: [],
                               combinator: nil}, _tree) do
     false
   end
@@ -38,7 +38,7 @@ defmodule Floki.Selector do
       && type_match?(html_node, selector.type)
       && classes_matches?(html_node, selector.classes)
       && attributes_matches?(html_node, selector.attributes)
-      && pseudo_class_match?(html_node, selector.pseudo_class, tree)
+      && pseudo_classes_match?(html_node, selector.pseudo_classes, tree)
   end
 
   defp id_match?(_node, nil), do: true
@@ -96,10 +96,11 @@ defmodule Floki.Selector do
     end
   end
 
-  defp pseudo_class_match?(_html_node, nil, _tree), do: true
-  defp pseudo_class_match?(html_node, pseudo_classes, tree) when is_list(pseudo_classes) do
-    Enum.all?(pseudo_classes, &(pseudo_class_match?(html_node, &1, tree)))
+  defp pseudo_classes_match?(_html_node, [], _tree), do: true
+  defp pseudo_classes_match?(html_node, pseudo_classes, tree) do
+    Enum.all?(pseudo_classes, &pseudo_class_match?(html_node, &1, tree))
   end
+
   defp pseudo_class_match?(html_node, pseudo_class, tree) do
     case pseudo_class.name do
       "nth-child" ->
@@ -107,10 +108,7 @@ defmodule Floki.Selector do
       "first-child" ->
         PseudoClass.match_nth_child?(tree, html_node, %PseudoClass{name: "nth-child", value: 1})
       "not" ->
-        case pseudo_class.value do
-          selectors when is_list(selectors) -> Enum.all?(selectors, &(!Selector.match?(html_node, &1, tree)))
-          selector -> !Selector.match?(html_node, selector, tree)
-        end
+        Enum.all?(pseudo_class.value, &(!Selector.match?(html_node, &1, tree)))
       "fl-contains" ->
         PseudoClass.match_contains?(tree, html_node, pseudo_class)
       unknown_pseudo_class ->
--- a/lib/floki/selector.ex
+++ b/lib/floki/selector.ex
@@ -12,7 +12,7 @@ defmodule Floki.Selector do
             classes: [],
             attributes: [],
             namespace: nil,
-            pseudo_class: nil,
+            pseudo_classes: [],
             combinator: nil
 
   @doc false
--- a/lib/floki/selector.ex
+++ b/lib/floki/selector.ex
@@ -97,6 +97,9 @@ defmodule Floki.Selector do
   end
 
   defp pseudo_class_match?(_html_node, nil, _tree), do: true
+  defp pseudo_class_match?(html_node, pseudo_classes, tree) when is_list(pseudo_classes) do
+    Enum.all?(pseudo_classes, &(pseudo_class_match?(html_node, &1, tree)))
+  end
   defp pseudo_class_match?(html_node, pseudo_class, tree) do
     case pseudo_class.name do
       "nth-child" ->
--- a/lib/floki/selector.ex
+++ b/lib/floki/selector.ex
@@ -104,7 +104,10 @@ defmodule Floki.Selector do
       "first-child" ->
         PseudoClass.match_nth_child?(tree, html_node, %PseudoClass{name: "nth-child", value: 1})
       "not" ->
-        !Selector.match?(html_node, pseudo_class.value, tree)
+        case pseudo_class.value do
+          selectors when is_list(selectors) -> Enum.all?(selectors, &(!Selector.match?(html_node, &1, tree)))
+          selector -> !Selector.match?(html_node, selector, tree)
+        end
       "fl-contains" ->
         PseudoClass.match_contains?(tree, html_node, pseudo_class)
       unknown_pseudo_class ->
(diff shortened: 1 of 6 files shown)


mischov:
CSS Selectors Level 3 only supports a single selector inside of the `:not()` pseudo class, so not supporting a list of selectors inside of `:not()` would be the correct behavior.

CSS Selectors Level 4 proposes supporting lists of selectors (and even complex selectors) inside of `:not()`, but as I understand it this behavior does not have widespread adoption. Lack of widespread adoption does not mean Floki should not support this behavior, but I think Floki has so far kept closer to CSS Selectors Level 3 (it explicitly disallows complex selectors in `:not()`).

Were Floki to support multiple pseudo selectors I would suggest `a.link:not(.bar):not([style=baz])`, but I don't think Floki does.

jjcarstens:
That's a good point @mischov. In most of my browser testing, selector lists were working just fine and checking support in each browser showed it _worked_, but not officially supported.

I just tested with `a:link:not(.foo):not(.bar)` and the selector is parsed, but doesn't filter out the results with anything past the first pseudo-class `:not`. I'll see if I can get both supported.

Personally, I think this should be fine supporting behavior from CSS Selectors Level 4. Folki also has some custom iterations of pseudo-class processing outside of standard scopes (i.e. `:fl-contains('foobar')` is a good example of the most recent implementation). As long as both the selector list and `:not():not()` are supported

mischov:
@jjcarstens 
Especially given that Floki already supports compound selectors in `:not()`, I agree with you that supporting additional Selectors Level 4 behavior should be fine.

I think the issue with multiple pseudo classes in Floki is that only a single slot for a pseudo class is provided.

jjcarstens:
Added support for multiple `:not` declarations:
`a.link:not(.class):not(#id)`

and also supports `a.link:not(.class, .class2):not(#id)`

philss:
@jjcarstens @mischov sorry guys, I'm a little bit busy lately.

@jjcarstens Thank you very much for this PR! 😃
I have some considerations:

- It would be better if we refactor the `Selector` and `SelectorParser` to treat `pseudo_class` as always`pseudo_classes`. Working only with lists would make the implementation easier, and would make more flexible as well. It's a thing I wanted to do before;
- I think we should add tests that check the behavior of `:not(.foo):not(.bar)` version. Can you add?
- I'm not very comfortable to add support for the version with list of selectors separated by comma that is proposed by [CSS4](https://www.w3.org/TR/selectors4/#negation). One of the reasons is what @mischov commented (the lack of adoption), and the other is that I think it made the implementation a little more complicated. I would prefer to refactor to not have that "split" of commas, and not have that regex too before introducing this feature. What do you think? Did you always used this way before?

I want to support some features of CSS4. This specific feature that I'm not comfortable with right now.

philss:
@jjcarstens please tell me if you think it can worth the cost I mentioned. We can also go with this and improve later if you think so. :)

PS: I said complicated but I meant complex.

lib/floki/selector.ex
```diff
@@ -97,14 +97,20 @@ defmodule Floki.Selector do
   end
 
   defp pseudo_class_match?(_html_node, nil, _tree), do: true
+  defp pseudo_class_match?(html_node, pseudo_classes, tree) when is_list(pseudo_classes) do
```
philss: Treating as lists would remove this guard, and the line above.

lib/floki/selector.ex
```diff
@@ -97,14 +97,20 @@ defmodule Floki.Selector do
...
   defp pseudo_class_match?(html_node, pseudo_class, tree) do
     case pseudo_class.name do
       "nth-child" ->
         PseudoClass.match_nth_child?(tree, html_node, pseudo_class)
       "first-child" ->
         PseudoClass.match_nth_child?(tree, html_node, %PseudoClass{name: "nth-child", value: 1})
       "not" ->
-        !Selector.match?(html_node, pseudo_class.value, tree)
+        case pseudo_class.value do
+          selectors when is_list(selectors) -> Enum.all?(selectors, &(!Selector.match?(html_node, &1, tree)))
```
philss: Removing the `comma` inside `:not()` would remove this `case`, right?

mischov:
> It would be better if we refactor the Selector and SelectorParser to treat pseudo_class as alwayspseudo_classes. Working only with lists would make the implementation easier, and would make more flexible as well. It's a thing I wanted to do before

That would be my suggestion as well.

> I'm not very comfortable to add support for the version with list of selectors separated by comma that is proposed by CSS4 ... I think it made the implementation a little more complicated. I would prefer to refactor to not have that "split" of commas, and not have that regex too before introducing this feature. What do you think? Did you always used this way before?

I don't *think* that removing the split and allowing a list of comma separated selectors needs to complicate things.

I added this functionality earlier today in Meeseeks, and all that was really involved in the process was adding an initial step to the sector parsing that handles the many-selectors case. In Floki `do_parse` takes tokens and an element as args, but you could also have a `do_parses` or something that would take tokens and an element accumulator as args.

This is what it looked like when I added it:
https://github.com/mischov/meeseeks/blob/82dd7bf77dc5fea083677626890c9ce4c16a7448/lib/meeseeks/selector/css/parser.ex#L16-L43

jjcarstens:
* I'll add those test cases. Totally slipped my mind

* I actually would prefer to treat it as `pseudo_classes`. I initially did it this way thinking it would prevent edge cases I hadn't thought through with existing code. Afterwards, I didn't fully like the result and thought I'd let you weigh in.

* I'm not totally attached to keeping the comma split. That's how I've tested in browsers and is probably more attributed to laziness and typing less characters. it's nice to have and I can investigate based off @mischov suggestion, but would be fine to drop it for now if that's not what you'd like to support. Either way, happy to make those changes now

philss:
@jjcarstens @mischov 
OK guys, let's add this feature! We can refactor later based on what @mischov suggested (to remove the `split`).

Thank you both!

lib/floki/selector_parser.ex
```diff
@@ -42,49 +42,33 @@ defmodule Floki.SelectorParser do
...
-    pseudo_class = selector.pseudo_class
-    do_parse(t, %{selector | pseudo_class: %{pseudo_class | value: to_string(pattern)}})
+    [pseudo_class | pseudo_classes] = selector.pseudo_classes
+    do_parse(t, %{selector | pseudo_classes: [%{pseudo_class | value: to_string(pattern)} | pseudo_classes]})
   end
   defp do_parse([{:pseudo_class_quoted, _, pattern} | t], selector) do
-    pseudo_class = selector.pseudo_class
-    do_parse(t, %{selector | pseudo_class: %{pseudo_class | value: to_string(pattern)}})
-  end
-  defp do_parse([{:pseudo_class_generic_value, _, value} | t], selector) do
```
jjcarstens: I couldn't find anywhere this was used and it seemed to be older, duplicate code so I went ahead and removed it in 9b2eee2. Let me know if it needs to stay in.

lib/floki/selector_parser.ex
```diff
@@ -157,28 +141,34 @@ defmodule Floki.SelectorParser do
...
+    pseudo_class = update_pseudo_not_value(pseudo_class, pseudo_not_selector)
+    do_parse_pseudo_not(t, pseudo_class)
+  end
+  defp do_parse_pseudo_not([{:space, _} | t], pseudo_not_selector, pseudo_class) do
+    do_parse_pseudo_not(t, pseudo_not_selector, pseudo_class)
   end
-  defp split_list_for_not_tokens([], tokens_before) do
-    {[], tokens_before}
+  defp do_parse_pseudo_not([next_token | t], pseudo_not_selector, pseudo_class) do
+    pseudo_not_selector = do_parse([next_token], pseudo_not_selector)
```
jjcarstens: This seemed cleaner to parse the next token one by one within the parenthesis of `:not()` building all selectors within before leaving the pseudo_not parsing. It also avoids having to change any `do_parse` logic and trying to handle quite a few different edge cases

lib/floki/selector_parser.ex
philss: Yes, it is not in use. I forgot to delete. 🔥

lib/floki/selector_parser.ex
philss: Nice!

philss:
@jjcarstens I think this is awesome! The possibility to use multiple pseudo-class selectors is great too!
We can think in a way to get rid of the Regex later. It's good enough for now!

Thank you very much! 💜

PS: I'm gonna squash into one commit to make the history more clean.

philss:
@jjcarstens @mischov this was released in version `0.17.0`. Thanks!

PR #106, opened 2017-04-13
Update selector parser to handle comma separated selectors

mischov:
This PR updates the selector parser to handle strings including comma separated selectors, and removes the string split in finder's `get_selectors` (as we discussed in #105).

Selector parser `parse` now returns a list of selectors instead of a single selector.


philss:
Great work, @mischov! Thank you! 😃

PR #107, opened 2017-05-03
Ignore style tag content in Floki.text/1

aaronrenner:
Currently if there is a `<style>` tag in the body of the html, its content is included when calling `Floki.text/1`. This omits the content of the style tag from the text output.

### DIFF
--- a/lib/floki/deep_text.ex
+++ b/lib/floki/deep_text.ex
@@ -21,6 +21,7 @@ defmodule Floki.DeepText do
   end
   defp get_text({:comment, _}, acc, _), do: acc
   defp get_text({"br", _, _}, acc, _), do: acc <> "\n"
+  defp get_text({"style", _, _}, acc, _), do: acc
   defp get_text({_, _, nodes}, acc, sep) do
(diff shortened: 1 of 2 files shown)


mischov:
So then how does somebody get the text content of the style tag if that is what you want?

My solution in Meeseeks was to omit the value of `<style>` and `<script>` tags from `Meeseeks.text` but to provide a `Meeseeks.data` extractor that returns those two values. Would that be a good solution here?

aaronrenner:
Good point. I'm new to floki and was looking at calling `Floki.text/1` on an entire page, and not a single tag. Does floki already have a mechanism to get the contents of a `<script>` tag?

mischov:
Text with the `js` option assigned to true.

aaronrenner:
As a new user to Floki, I don't think I'm familiar enough with the library to make decisions that affect the public API. This PR is just to address some behavior that seemed unexpected. Maybe this is a larger issue than I realized. If you or @philss have suggestions of what the public API should look like, I'd be happy to work on a new PR.

PR #108, opened 2017-05-05
Fix failing finding on unescaped angles in text

sergey-kintsel:
I found that issue during parsing `http://ruby-doc.org/stdlib-2.4.1/libdoc/psych/rdoc/Psych/Parser.html`.
It has this span on it


philss:
@sergey-kintsel Thank you! 😃

philss:
I forgot to mention that this fix was released in version `0.17.1`.

PR #109, opened 2017-05-13
Extend Travis build matrix

samhamilton:
Hi, 

A small update to always test the latest Elixir patch versions and also bump the OTP version. 

Thanks
Sam


philss:
@samhamilton Thank you! 😃

PR #111, opened 2017-05-20
fixed crash caused by xml declaration inside body

erikdsi:
I was trying to parse a page that for some unknown reason has a xml declaration tag lost inside the body tag that caused Floki to crash.
I added a test for that edge case and a quick fix that just ignores it.

### DIFF
--- a/lib/floki/html_tree.ex
+++ b/lib/floki/html_tree.ex
@@ -99,14 +99,18 @@ defmodule Floki.HTMLTree do
   defp build_tree(tree, [], _, []), do: tree
   defp build_tree(tree, [{tag, attrs, child_children} | children], parent_id, stack) do
     new_id = IDSeeder.seed(tree.node_ids)
-    new_node = %HTMLNode{type: tag,
-                         attributes: attrs,
-                         node_id: new_id,
-                         parent_node_id: parent_id}
+    nodes = if tag != :pi do
+      new_node = %HTMLNode{type: tag,
+                           attributes: attrs,
+                           node_id: new_id,
+                           parent_node_id: parent_id}
 
-    nodes = put_new_node(tree.nodes, new_node)
+      put_new_node(tree.nodes, new_node)
+    else
+      tree.nodes
+    end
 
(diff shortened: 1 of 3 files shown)


lib/floki/html_tree.ex
```diff
@@ -99,14 +99,18 @@ defmodule Floki.HTMLTree do
   defp build_tree(tree, [], _, []), do: tree
   defp build_tree(tree, [{tag, attrs, child_children} | children], parent_id, stack) do
     new_id = IDSeeder.seed(tree.node_ids)
-    new_node = %HTMLNode{type: tag,
-                         attributes: attrs,
-                         node_id: new_id,
-                         parent_node_id: parent_id}
+    nodes = if tag != :pi do
```
philss: I think we could ignore the nodes that are `:pi` in a function match.
So, above this function declaration, we could have a declaration more or less like this:

```elixir
defp build_tree(tree, [{:pi, _, _} | children], parent_id, stack), do: build_tree(tree, children, parent_id, stack)
```

This version would avoid an `if`. What do you think?

sabondano: Hey, I was running into similar issues. Wondering if you know what "pi" stands for? Just curious.

erikdsi: I implemented, tested, works and is definitely a more idiomatic way to solve it.

lib/floki/html_tree.ex
```diff
@@ -97,16 +97,18 @@ defmodule Floki.HTMLTree do
   defp get_ids_for_delete_stack(_), do: []
 
   defp build_tree(tree, [], _, []), do: tree
+  defp build_tree(tree, [{:pi, _, _} | children], parent_id, stack), do: build_tree(tree, children, parent_id, stack)
```
sabondano: @erikdsi do you know what "pi" stands for?

mischov: processing instruction

sabondano: Thank you :)

philss:
@erikdsi Thank you! 😃

philss:
I forgot to mention earlier: this fix is available in version `0.17.1` and above.

PR #112, opened 2017-05-20
Fix issue when finding fails if html begins with xml tag

sergey-kintsel:


philss:
Hi @sergey-kintsel, thank you for this PR.

I'm not sure if this is the same issue that was fixed by #111 . Can you check, please?
Also, if this issue is different, please update your branch with master.

sergey-kintsel:
Hi @philss! I rebased on master, commented my crutch-line in `selector.ex` and my spec failed. So, I assume that this is needed :-)

philss:
@sergey-kintsel OK, thank you! 😃

philss:
I forgot to mention that this fix was released in version 0.17.1.

Issue #110, opened 2017-05-14
FunctionClauseError for :not pseudo selector and equal attribute value check


Eiji7:
Example code:
```elixir
iex> html = "<select id='my_select'><option value=''></option><option value='1'>1</option></select>"
"<select id='my_select'><option value=''></option><option value='1'>1</option></select>"
iex> Floki.find(html, "#my_select > option:not([value=''])")  
** (FunctionClauseError) no function clause matching in Floki.SelectorParser.do_parse/2
    (floki) lib/floki/selector_parser.ex:31: Floki.SelectorParser.do_parse([equal: 1], %Floki.Selector{attributes: [%Floki.AttributeSelector{attribute: nil, match_type: nil, value: nil}], classes: [], combinator: nil, id: nil, namespace: nil, pseudo_classes: [], type: "value"})
    (floki) lib/floki/selector_parser.ex:171: Floki.SelectorParser.do_parse_pseudo_not/3
    (floki) lib/floki/selector_parser.ex:54: Floki.SelectorParser.do_parse/2
    (floki) lib/floki/selector_parser.ex:145: Floki.SelectorParser.consume_combinator/2
    (floki) lib/floki/selector_parser.ex:88: Floki.SelectorParser.do_parse/2
    (floki) lib/floki/selector_parser.ex:27: Floki.SelectorParser.do_parse_all/2
    (floki) lib/floki/finder.ex:24: Floki.Finder.find/2
    (floki) lib/floki.ex:146: Floki.find/2
```
I tested on latest `0.17.0` release and `master` branch - both returns same exception to console.
**Important**: this example works in version: `0.16.0`!
ping @philss

philss:
@Eiji7 Thank you for the report. I'm looking into it.

jjcarstens:
Tested this with #114 and should be golden once that is merged:
```elixir
iex> html = "<select id='my_select'><option value=''></option><option value='1'>1</option></select>"
iex> Floki.find(html, "#my_select > option:not([value=''])")
[{"option", [{"value", "1"}], ["1"]}]
```

PR #114, opened 2017-05-24
Handle attribute selectors in :not()

jjcarstens:
Simple update to allow pseudo `:not()` selects to handle attribute selectors within the parenthesis:
`a.link:not([class*=baz])` etc etc

@philss 

(ps. I'm not sure if you want mix.lock, but it looks like some deps were updated it so I included. Can drop easily if needed)

### DIFF
--- a/lib/floki/selector_parser.ex
+++ b/lib/floki/selector_parser.ex
@@ -29,6 +29,7 @@ defmodule Floki.SelectorParser do
   end
 
   defp do_parse([], selector), do: {selector, []}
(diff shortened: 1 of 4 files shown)


jjcarstens:
resolves https://github.com/philss/floki/issues/110

Eiji7:
@jjcarstens: I checked both example code and private project that uses `Floki` and both works, good job!

philss:
@jjcarstens awesome! Thank you very much! 💜

philss:
This fix was released in version `0.17.2`.

PR #115, opened 2017-05-24
Return selector after consuming combinator

jjcarstens:
Since [the change to parse all tokens, a selector is returned when encountering a `:comma` token](https://github.com/philss/floki/pull/106/files#diff-7f26dd84090995cad0112f38a801b18bR32). 

When we `consume_combinators`, that calls `do_parse` which gets a selector and continues on parsing tokens. However, the selector is returned as a result of matching `:comma` token and removes it from the remaining tokens. So when parsing continues after `consume_combinator`, it continues parsing the previous selector instead of starting a new one.

This changes so `consume_combinator` functions return a selector tuple since they will get a fully completed selector after parsing.

```elixir
# before the change
iex> "a b.foo c, ul a[class=bar]" |> Floki.SelectorTokenizer.tokenize |> Floki.SelectorParser.parse |> length
1
iex> "a b.foo c, ul a[class=bar]" |> Floki.SelectorTokenizer.tokenize |> Floki.SelectorParser.parse
[%Floki.Selector{attributes: [], classes: [],
  combinator: %Floki.Combinator{match_type: :descendant,
   selector: %Floki.Selector{attributes: [], classes: ["foo"],
    combinator: %Floki.Combinator{match_type: :descendant,
     selector: %Floki.Selector{attributes: [%Floki.AttributeSelector{attribute: "class",
        match_type: :equal, value: "bar"}], classes: [], combinator: nil,
      id: nil, namespace: nil, pseudo_classes: [], type: "a"}}, id: nil,
    namespace: nil, pseudo_classes: [], type: "ul"}}, id: nil, namespace: nil,
  pseudo_classes: [], type: "a"}]

# With proposed changes
iex> "a b.foo c, ul a[class=bar]" |> Floki.SelectorTokenizer.tokenize |> Floki.SelectorParser.parse |> length
2
iex> "a b.foo c, ul a[class=bar]" |> Floki.SelectorTokenizer.tokenize |> Floki.SelectorParser.parse
[%Floki.Selector{attributes: [], classes: [],
  combinator: %Floki.Combinator{match_type: :descendant,
   selector: %Floki.Selector{attributes: [], classes: ["foo"],
    combinator: %Floki.Combinator{match_type: :descendant,
     selector: %Floki.Selector{attributes: [], classes: [], combinator: nil,
      id: nil, namespace: nil, pseudo_classes: [], type: "c"}}, id: nil,
    namespace: nil, pseudo_classes: [], type: "b"}}, id: nil, namespace: nil,
  pseudo_classes: [], type: "a"},
 %Floki.Selector{attributes: [], classes: [],
  combinator: %Floki.Combinator{match_type: :descendant,
   selector: %Floki.Selector{attributes: [%Floki.AttributeSelector{attribute: "class",
      match_type: :equal, value: "bar"}], classes: [], combinator: nil, id: nil,
    namespace: nil, pseudo_classes: [], type: "a"}}, id: nil, namespace: nil,
  pseudo_classes: [], type: "ul"}]
```

I _think_ this covers all the missing cases, but would definitely love close 👀 and any suggestions you believe may still need to be tested.

### DIFF
--- a/lib/floki/selector_parser.ex
+++ b/lib/floki/selector_parser.ex
@@ -80,24 +80,24 @@ defmodule Floki.SelectorParser do
     do_parse(t, %{selector | pseudo_classes: [%{pseudo_class | value: to_string(pattern)} | pseudo_classes]})
   end
   defp do_parse([{:space, _} | t], selector) do
-    {t, combinator} = consume_combinator(t, :descendant)
+    {remaining_tokens, combinator} = consume_combinator(t, :descendant)
 
-    do_parse(t, %{selector | combinator: combinator})
+    {%{selector | combinator: combinator}, remaining_tokens}
   end
   defp do_parse([{:greater, _} | t], selector) do
-    {t, combinator} = consume_combinator(t, :child)
+    {remaining_tokens, combinator} = consume_combinator(t, :child)
 
-    do_parse(t, %{selector | combinator: combinator})
+    {%{selector | combinator: combinator}, remaining_tokens}
   end
   defp do_parse([{:plus, _} | t], selector) do
-    {t, combinator} = consume_combinator(t, :sibling)
+    {remaining_tokens, combinator} = consume_combinator(t, :sibling)
 
-    do_parse(t, %{selector | combinator: combinator})
+    {%{selector | combinator: combinator}, remaining_tokens}
   end
   defp do_parse([{:tilde, _} | t], selector) do
-    {t, combinator} = consume_combinator(t, :general_sibling)
+    {remaining_tokens, combinator} = consume_combinator(t, :general_sibling)
 
-    do_parse(t, %{selector | combinator: combinator})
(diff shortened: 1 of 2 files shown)


mischov:
Good catch. Your changes should be correct.

In [the code that #106 is based on](https://github.com/mischov/meeseeks/blob/2234e0e4fb87090f266a3887fb81bbcfef8fb8af/lib/meeseeks/selector/css/parser.ex#L104) I append a comma onto the tokens returned from consuming a combinator, forcing a return of the selector in the next step (because reasons).

The behavior of appending a comma to the tokens was overkill in Floki so I removed it but I (clearly) forgot to change the code to return the `{remaining_tokens, selector}` tuple instead of continuing to parse the element.

philss:
I think this is correct! Thank you again, @jjcarstens!

philss:
This fix was released in version `0.17.2`

PR #121, opened 2017-07-03
Fix attr function when the attribute is absent

sergey-kintsel:
If it is absent the mutation function is being called with `nil`. This
allows to add an attribute.


philss:
@sergey-kintsel Thank you!

PR #123, opened 2017-07-04
Fix converting html_tree to raw html when it has xml prefix

sergey-kintsel:
Previously it was failing with:
(Protocol.UndefinedError) protocol Enumerable not implemented for "xml"


philss:
@sergey-kintsel Thank you again! 😃

PR #126, opened 2017-07-14
I bet you'll forgot version number in the README again

asiniy:


philss:
Hi @asiniy! Thank you for the PR, and sorry for the delay.
I think we should have the most recent version on `README.md`. I usually use the last `MINOR` version with the `~>` operator, since it will always install the most recent version of `PATCH`.

So I will decline this PR in order to keep updating the `README.md`. Thank you again.

Issue #118, opened 2017-06-15
Possibly rename Floki.apply_transformation to Floki.map


aphillipo:
Hi @philss! Thanks again for Floki... 

I have a quick question about apply_transformation - in my head I was looking for Floki.map rather than Floki.apply_transformation.  I wonder if you think the current naming is correct?

In my head Enum.map means iterate over all all items in an Enumerable; Floki.map means the same for all elements in a Floki html_tree?

I'm going to use apply_transformation now because I need to blur out some text and scramble the letters before payment unblurs them with unscrambled text.

I'm happy to rename it and add some examples if you want a pull request?  Thanks!

philss:
Hi @aphillipo! 

I agree with you. I'm not that happy with the current name, but I did not realise that this is basically a map. I would love if you rename it!

I also think in the future we could add an implementation of the `Enumerable` protocol to the `Floki.HTMLTree` struct. This would make easy to traverse and apply some transformations.

Thank you!

aphillipo:
Oh!  That is a great idea @philss - would all of Enum then work with Floki?

aphillipo:
I'm looking at implementing Enumerable for ```Floki.HTMLTree``` now.  It'll be good for to level up my Elixir ability!  Thanks!

philss:
Yes, I never did such thing (implementing a protocol for a complex struct), but I think it works! Ping me if you need help with something. Thanks!

aphillipo:
I'm not certain I don't prefer the original tuples/lists of the html_tree type :-/

Something like this get's us 99% of the way there, but I'm not sure you can implement Enumerable unless you have a defstruct type.

https://github.com/rrrene/html_sanitize_ex/blob/master/lib/html_sanitize_ex/traverser.ex

A little confused by the ordering of Floki.HTMLTree:

1) Root Node looks to always be 1, but also it seems possible to have multiple root nodes? Go through these first. floki_tree[1], floki_tree[2] etc.

2) Each time child nodes has values [h|t] add the head to the accumulator with {:cont} as the Enumerable action and recurse on the tail.  Will probably get the actual nodes here for all children and pass them in.

3) Can probably use the default implementations for count and member (for now) because they just use the reducer.  Presumably the match for member would be an exact match on the particular struct.

4) Feel like writing this out have helped a bit.  Liking the HTMLTree a bit more now.  Let me try again!

philss:
Hi @aphillipo!

Answering the points:

[...]

PR #128, opened 2017-07-22
Renames Floki.transform/2 and Floki.Finder#apply_transformation/2

philss:
It renames these transformation function to only `map/2` following
@aphillipo's suggestion from https://github.com/philss/floki/issues/118


philss:
It closes #118.

PR #146, opened 2017-10-21
Add :last-child pseudo-class selector

### LINKED ISSUE #143: Add :last-child pseudo-class selector
This pseudo-class selector returns the node that is the last-child of a given parent. You can check examples here: https://developer.mozilla.org/en-US/docs/Web/CSS/:last-child

The ideia here is to work with [`Floki.Selector.PseudoClass`](https://github.com/philss/floki/blob/master/lib/floki/selector/pseudo_class.ex) module, and with the branch that is present on `pseudo_class_match?/3` (private function) that is present on [`Floki.Selector`](https://github.com/philss/floki/blob/master/lib/floki/selector.ex#L129-L145).

seleem1337:
Hello,
This PR closes #143 by adding support for the pseudo-class selector `:last-child`. A matching branch is added to `pseudo_class_match?/3` in `Floki.Selector`. Also, the `node_position/2` in `Floki.Selector.PseudoClass` is now split to `node_position/2` and `get_children_nodes_ids/2`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -198,6 +198,7 @@ Here you find all the [CSS selectors](https://www.w3.org/TR/selectors/#selectors
 | E[foo\|="en"]    | an E element whose "foo" attribute has a hyphen-separated list of values beginning (from the left) with "en" |
 | E:nth-child(n)  | an E element, the n-th child of its parent |
 | E:first-child   | an E element, first child of its parent |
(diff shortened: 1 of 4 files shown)


philss:
Awesome, @seleem1337! Thank you! 😃

Issue #142, opened 2017-10-21
Format codebase files with the upcoming code formatter


philss:
This is a copy of the issue that was created by José at Elixir's core repository: https://github.com/elixir-lang/elixir/issues/6643. So please read that issue before proceeding.

The idea is to run the upcoming Elixir formatter against all modules. To encourage more people to contribute, I will ask you to format only one module (and its test module) per pull request.

Please open the pull request with the following title format:

`Run the code formatter on Module.Name.Here`

philss:
closed by #152

PR #148, opened 2017-10-21
Format test

fcevado:
#142

### DIFF
--- a/lib/.#floki.ex
+++ b/lib/.#floki.ex
@@ -1,1 +1,0 @@
-flavio@j2l2r32.remote.office.telnyx.com.8526:1508604136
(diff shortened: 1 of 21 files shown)


lib/.#floki.ex
```diff
@@ -0,0 +1 @@
+flavio@j2l2r32.remote.office.telnyx.com.8526:1508604136
```
philss: I think this file was added accidentally. Can you remove it?

fcevado:
i guess i messed with something when resolving conflicts... I'll take a look later today, but i guess it will work better if do it from the begining... and probably will need to address that too, to work with Elixir 1.6:
https://github.com/elixir-lang/elixir/issues/6955

PR #149, opened 2017-10-21
Add :nth-of-type(n), :first-of-type, and :last-of-type pseudo-class selectors

### LINKED ISSUE #144: Add :first-of-type pseudo-class selector
This pseudo-class selector will return only the first node of a given type (or any type, if universal selector - `*` is used). Examples: https://developer.mozilla.org/en-US/docs/Web/CSS/:first-of-type

The area related to this feature is very similar to the issue https://github.com/philss/floki/issues/143.

seleem1337:
Hello,
This PR closes #144 by adding support for the pseudo-class selectors: `:nth-of-child(n)`, `:first-of-type` and `:last-of-type`.

### DIFF
--- a/lib/floki/selector/pseudo_class.ex
+++ b/lib/floki/selector/pseudo_class.ex
@@ -47,7 +47,7 @@ defmodule Floki.Selector.PseudoClass do
 
   def match_nth_of_type?(_, %HTMLNode{parent_node_id: nil}, _), do: false
   def match_nth_of_type?(tree, html_node, %__MODULE__{value: -1}) do
-    children_nodes_ids = get_children_nodes_indexed_by_type(tree, html_node.parent_node_id, html_node.type)
+    children_nodes_ids = get_children_nodes_indexed_by_type(tree, html_node)
     {last_child_id, _} = Enum.max_by(children_nodes_ids, fn({_, pos}) -> pos end)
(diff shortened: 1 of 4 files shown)


lib/floki/selector/pseudo_class.ex
```diff
@@ -45,6 +45,31 @@ defmodule Floki.Selector.PseudoClass do
     false
   end
 
+  def match_nth_of_type?(_, %HTMLNode{parent_node_id: nil}, _), do: false
+  def match_nth_of_type?(tree, html_node, %__MODULE__{value: -1}) do
+    children_nodes_ids = get_children_nodes_indexed_by_type(tree, html_node.parent_node_id, html_node.type)
```
philss: Since you are accessing two attributes of the same record, I would recommend to only pass the `html_node` to the `get_children_nodes_indexed_by_type` function (changing the arity from 3 to 2).

seleem1337: I initially passed the `parent_node_id` and `type` separately to be consistent with `get_children_nodes_indexed/2` but you're right, it makes sense to pass `html_node` as a whole. I've pushed a commit which should address this. Thanks.

philss:
Awesome work, @seleem1337! Thank you again! 😄

PR #150, opened 2017-10-22
Add Functional Notation nth-child selector

### LINKED ISSUE #145: Add complex selections to `:nth-child` pseudo-class selector
Today we only accept simple selectors, like `:nth-child(1)`, `:nth-child(odd)` or `:nth-child(even)`.

The idea is to support more complex selectors, like: `:nth-child(2n+1)`.

Examples can be found here: https://developer.mozilla.org/en-US/docs/Web/CSS/:nth-child

nirev:
https://developer.mozilla.org/en-US/docs/Web/CSS/:nth-child#Functional_notation

Closes #145

### DIFF
--- a/lib/floki/selector/functional.ex
+++ b/lib/floki/selector/functional.ex
@@ -1,0 +1,44 @@
+defmodule Floki.Selector.Functional do
+  @moduledoc false
+  # Represents a functional notation for a selector
+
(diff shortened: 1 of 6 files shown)


philss:
Awesome, @nirev! 😃 
Thank you!

Issue #156, opened 2017-11-27
Test with html5ever HTML parser on CI


philss:
Today we only test Floki with the default parser, which is Mochiweb's parser.

We need to test with `html5ever` too in order to prevent bugs without noticing from our tests.

This would avoid the problem on issue #155.

PR #157, opened 2017-11-29
Handle doctype in raw_html, fixes #155

### LINKED ISSUE #155: Doctype triggers no function clause matching raw_html/2
Sending HTML with a doctype through `parse` and `raw_html` produces `** (FunctionClauseError) no function clause matching in Floki.raw_html/2`

```elixir
# Does not work
"<!doctype html><html><head><title>Test</title></head><body>Body</body></html>"
|> Floki.parse() 
|> Floki.raw_html()

# Works as expected (returning original string)
"<html><head><title>Test</title></head><body>Body</body></html>"
|> Floki.parse() 
|> Floki.raw_html()
```

Inspecting into the input it looks like after parse the first tuple is `{:doctype, "html", "", ""}` which has four elements and the general match is only expecting three [`defp raw_html([{type, attrs, children}|tail], html)`](https://github.com/philss/floki/blob/v0.19.0/lib/floki.ex#L121)?

jhchen:
Not sure how to add tests but the cases I tried were

`<!DOCTYPE html>`
`<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.01//EN" "http://www.w3.org/TR/html4/strict.dtd">`
`<!DOCTYPE note SYSTEM "Note.dtd">`

using `Floki.parse()` and `Floki.raw_html()`

### DIFF
--- a/lib/floki.ex
+++ b/lib/floki.ex
@@ -118,6 +118,16 @@ defmodule Floki do
     raw_html(tail, html <> "<?xml " <> tag_attrs(attrs) <> "?>")
   end
 
+  defp raw_html([{:doctype, type, public, system} | tail], html) do
+    attr =
+      case {public, system} do
+        {"", ""} -> ""
+        {"", system} -> " SYSTEM \"#{system}\""
+        {public, system} -> " PUBLIC \"#{public}\" \"#{system}\""
+      end
+    raw_html(tail, html <> "<!DOCTYPE #{type}#{attr}>")
+  end
+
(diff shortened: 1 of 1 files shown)


philss:
Thank you, @jhchen! 😃 

I added a test with this commit f379af3be7669a83f08768e080cc5a3889c55652.
We could test more scenarios later.

Also, I need to think about how to test those things that are different between the two parsers (https://github.com/philss/floki/issues/156)

PR #162, opened 2018-01-18
Updated specs for Floki.parse/1

grych:
`Floki.parse/1` returns binary when the string is not html, so the specs should cover it. Alternatively, you could change the `html_tree` type.


philss:
Thank you, @grych!

PR #168, opened 2018-01-26
fix some credo issues

davydog187:
There are still some consistency issues, but we can fix those with the formatter (which I will do in another PR if thats cool).

### DIFF
--- a/lib/floki/raw_html.ex
+++ b/lib/floki/raw_html.ex
@@ -1,7 +1,5 @@
 defmodule Floki.RawHTML do
-  @moduledoc"""
-  Converts html parse trees back into binary html
-  """
+  @moduledoc false
 
   @self_closing_tags [
     "area",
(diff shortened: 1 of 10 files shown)


lib/floki/raw_html.ex
```diff
@@ -1,4 +1,8 @@
 defmodule Floki.RawHTML do
+  @moduledoc"""
```
philss: I prefer not to document modules that are "private" for now. I think the the docs from the "public" function are enough: https://hexdocs.pm/floki/Floki.html#raw_html/1
Could you please check it to `@moduledoc false`?

davydog187:
@philss went ahead and ran the formatter as part of this PR, also addressed the remaining consistency issues.

davydog187:
It would be helpful if ebert was configured to run `mix format --check-fromatted` and `mix credo`

philss:
> It would be helpful if ebert was configured to run mix format --check-fromatted and mix credo

I think I can add this to the build, and fail if it is not following the format. Not sure if I can integrate with credo. If you know how, please go ahead with this too.

Thank you! 💜

Issue #161, opened 2018-01-09
raw_html encodes HTML where it should not


grych:
#158 fixed the issue with re-encoding HTML entities, but introduced the new issue I am facing in my project: it encodes them where it shouldn't:

Example:

```elixir
iex(1)> Floki.parse("<script>console.log(\"text\");</script>") |> Floki.raw_html  
"<script>console.log(&quot;text&quot;);</script>"

```
Expected (this is what we can observer in 0.18):

```elixir
iex(2)> Floki.parse("<script>console.log(\"text\");</script>") |> Floki.raw_html  
"<script>console.log(\"text\");</script>"
```

grych:
Thanks!
BTW it is not only an issue with `<script>`. 

```elixir
iex(5)> Floki.parse("<a>\"in a\"</a>") |> Floki.raw_html 
"<a>&quot;in a&quot;</a>"

```

Shouldn't `raw_html/1` preserve the text in the tag?

davydog187:
@philss I think this bug is more pervasive than what you're suggesting. This behavior is both inconsistent as well as incorrect. 

Inline `<style>` tags also get their inner css styles escaped, which can break css rules. For example

```elixir
iex(5)> text = "<style>.myclass[data-attr=\"stuff\"] { width: 100%; }</style>"
"<style>.myclass[data-attr=\"stuff\"] { width: 100%; }</style>"
iex(6)> Floki.parse(text) |> Floki.raw_html
"<style>.myclass[data-attr=&quot;stuff&quot;] { width: 100%; }</style>"
```

The other problem I see is that the intention of #159 was to keep input/output consistent, but its still breaking it when you don't want values escaped.

davydog187:
Can I propose we add second argument to `raw_html/1` that allows us to configure whether we escape or not?

Related, right now Floki exposes `raw_html/2`, which is an implementation detail, and should not be exposed to the user. @philss @grych if you agree with this approach I'd be happy to put in a PR

davydog187:
Another alternative is to note escape the insides of `<style>` or `<script>` as you mentioned, but I fear that there may be more use cases than what we're currently considering

grych:
Cool, configurable escape would solve my issue.

philss:
Hi @davydog187! Thank you for the input.

[...]

PR #169, opened 2018-01-26
configurable raw_html/2

davydog187:


lib/floki.ex
```diff
@@ -86,7 +86,7 @@ defmodule Floki do
 
   @spec raw_html(html_tree | binary) :: binary
```
davydog187: i need to update the typespec

lib/floki/raw_html.ex
```diff
@@ -18,35 +18,50 @@ defmodule Floki.RawHTML do
...
-  defp raw_html(string, _html) when is_binary(string), do: HtmlEntities.encode(string)
-  defp raw_html(tuple, html) when is_tuple(tuple), do: raw_html([tuple], html)
+  @doc """
+  """
+  def raw_html(html_tree, options) do
+    encoder =
+      case Keyword.fetch(options, :encode) do
+        {:ok, true} -> &HtmlEntities.encode/1
+        {:ok, false} -> &(&1)
+        :error -> &HtmlEntities.encode/1
```
davydog187: thoughts on making the default configurable at the Application level?

lib/floki/raw_html.ex
```diff
@@ -70,11 +85,11 @@ defmodule Floki.RawHTML do
...
   defp build_attrs(attr, attrs), do: "#{attrs} #{attr}"
 
-  defp tag_for(type, attrs, content) when type in ["script", "style"] do
+  defp tag_for(type, attrs, content, _encoder) when type in ["script", "style"] do
     tag_with_attrs(type, attrs, []) <> Enum.join(content) <> "</#{type}>"
   end
 
-  defp tag_for(type, attrs, children) do
-    tag_with_attrs(type, attrs, children) <> raw_html(children) <> close_end_tag(type, children)
+  defp tag_for(type, attrs, children, encoder) do
```
davydog187: I don't love the 4 arity, but it seemed cleaner than shoving everything inside `build_raw_html`

davydog187:
I'll add some tests as well

lib/floki/raw_html.ex
```diff
@@ -18,35 +18,50 @@ defmodule Floki.RawHTML do
     "wbr"
   ]
 
-  def raw_html(html_tree), do: raw_html(html_tree, "")
-  defp raw_html([], html), do: html
-  defp raw_html(string, _html) when is_binary(string), do: HtmlEntities.encode(string)
-  defp raw_html(tuple, html) when is_tuple(tuple), do: raw_html([tuple], html)
+  @doc """
```
davydog187: remove

lib/floki.ex
```diff
@@ -84,9 +84,9 @@ defmodule Floki do
 
   """
```
davydog187: add options to the docs

lib/floki/raw_html.ex
philss: This would be a nice to have feature too. It could fetch from `Application.get_env(:floki, :encode_raw_html)` on the `:error` clause of this `case`. Would be nice if we document that option too.

lib/floki.ex
philss: Can you add an example of the `encode: false` here too?

davydog187:
@philss addressed your feedback. Happy to change any language in the docs

philss:
@davydog187 Awesome, thank you!

philss:
PS: I will squash into one commit to make more easier to read the log.

davydog187:
This should have fixed #161

PR #170, opened 2018-01-30
Configure ebert for credo, run format

davydog187:
Added an ebert config, as it seems that credo wasn't running prior. I also ran the formatter once more.

May I suggest that we move this project to something along the lines of https://travis-ci.org/ ? Ebert seems pretty limited

### DIFF
--- a/.ebert.yml
+++ b/.ebert.yml
@@ -1,0 +1,3 @@
+engines:
+  credo:
+    enabled: true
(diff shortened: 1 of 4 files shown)


philss:
Thank you, @davydog187! I think we can integrate the run of Credo with "travis-ci".  We could also run "mix format --check-formatted" to make the build crash if it is not following the formatter.

davydog187:
Great, I'll look into it. Thanks @philss

PR #184, opened 2018-05-14
Fix style and script tags with comments

francois2metz:
In case of an html comment inside `style` or `script` tags, `Floki.raw_html` raise an error `(Protocol.UndefinedError) protocol String.Chars not implemented for {:comment, "...."}`.

So instead of reimplementing the serialization, I removed the default encoder to not encode html entities.


philss:
@francois2metz nice catch! Thank you!

Issue #191, opened 2018-10-09
Run the formatter on build


philss:
We should make the formatter run on each build, by using the `--check-formatted` flag.
This change will ensure that each pull request will fail if have formatting issues.

Examples:
- https://github.com/elixir-lang/flow/blob/master/.travis.yml#L24
- https://github.com/elixir-lang/gen_stage/blob/master/.travis.yml#L13

Notice that the [`formatter`](https://medium.com/blackode/code-formatter-the-big-feature-in-elixir-v1-6-0-f6572061a4ba) was introduced in Elixir 1.6. Therefore, we don't need to run it for versions lower than that.

hlhr:
I will submit a PR for this :)

hlhr:
I will submit a PR for this.

philss:
Closed by #195.

PR #195, opened 2018-10-09
Run the formatter on build

hlhr:
Fix for Issue #191 

Add script step to execute formatter (if version is > 1.5) and execute the tests in build pipeline


philss:
@hlhr Thank you again 😃 💚 !

PR #197, opened 2018-11-29
Add style block skiped in text

Vict0rynox:
Add opts `style` for skip text in <style> block.  (By default is true, for backward compatibility.)


philss:
@Vict0rynox awesome! Thank you! 💜

PR #200, opened 2019-01-30
Readme typo

thiamsantos:
Seems like we have a small typo on the readme. Instead of `html5ever` was written `http5ever`.


philss:
@thiamsantos Thank you! 💜

PR #201, opened 2019-01-31
Update html5ever version in README

lau:
Update html5ever version in the readme to the newest one. It is compatible with Erlang/OTP 21 unlike the previous version.


philss:
@lau Thank you! 💜

Issue #37, opened 2015-10-29
Create a built in HTML parser


philss:
Floki needs a HTML parser built in, in order to remove the [mochiweb](https://github.com/mochi/mochiweb) dependency. This will enable more flexibility and better control of the parsing step.

The parser goals are:
- [ ] support HTML5;
- [ ] support HTML snippets;
- [ ] be able to parse large files, like 15MB;
- [ ] easy to traverse;
- [ ] be a bit tolerant with errors, like missing closing tags.

philss:
Here is a test case with an example of error that Floki does not support today: https://github.com/henrik/sipper/commit/49a4c09afa8773f9253401608f89c8d1545124cf

Thanks @henrik for the example!

gmile:
@philss creating an html parser from scratch sounds like a huge amount of work. Have you thought about using a C library instead, such as this one https://github.com/google/gumbo-parser?

philss:
@gmile yeah, I thought about that, but what I want is to not depend on an external dependency.
This came from a bit of frustration with the Nokogiri ruby gem. It uses libxml2 and FFI to make the bridge. It failed so many times to compile with me that I didn't like the experience.

But, this is not discarded. I also think [Servo's HTML](https://github.com/servo/html5ever) is a good option.

gmile:
> But, this is not discarded

That said, are you specifically looking forward the Servo's HTML? Otherwise, I could play with gumbo-parser integration and see how it goes.

philss:
@gmile I'm not looking into this right now. So, please go for it. 👍

baron:
I was wondering what the expected behavior of a native html parser would be. Right now mochiweb_html.parse always returns empty lists in either the middle or the end (depending on what level of nesting the html has). I'm not sure if this is a bug or feature but it was confusing when I first started using the library because I was hoping for some kind of "to_hash" like function in ruby.

`iex(33)> htm = """
...(33)> <ul>
...(33)> <li>fooo</li>
...(33)> <li>bar</li>
...(33)> </ul>
...(33)> """
"<ul>\n<li>fooo</li>\n<li>bar</li>\n</ul>\n"
iex(34)> :mochiweb_html.parse(htm)
{"ul", [], [{"li", [], ["fooo"]}, {"li", [], ["bar"]}]}`

Would a replacement function recreate this behavior for backwards compatibility or break the api?

BTW, thanks for the awesome library!

Eiji7:
It would be awesome to have something like this:
[...]

PR #204, opened 2019-04-03
WIP: Internal tokenizer

philss:
This is work in progress of a HTML tokenizer.

It is related with https://github.com/philss/floki/issues/37


lib/floki/html/tree_construction.ex
```diff
@@ -0,0 +1,63 @@
+defmodule Floki.HTML.TreeConstruction do
```
mischov: Been a long road to get here- congrats!

Now the real "fun" starts. 😄

philss: @mischov thank you! 💜 😄  
I hope I can have some fun with all this complexity 😅

PR #208, opened 2019-06-10
Add Floki.traverse_and_update/2. Fix #206.

### LINKED ISSUE #206: Children/content manipulation support
Currently we can manipulate tag name and attributes with `Floki.map/2`, what about the children/content?

Use case is to override certain elements:
- Input: `picture` 
- Output: `<div>picture</div>`

ericlathrop:
Let me know if there's anything you want me to change/add.

### DIFF
--- a/lib/floki.ex
+++ b/lib/floki.ex
@@ -229,6 +229,22 @@ defmodule Floki do
   def map(html_tree, fun), do: Finder.map(html_tree, fun)
 
   @doc """
+  Traverses a HTML tree structure and returns a new tree structure that is the result of executing a function on all nodes. The function receives a tuple with {name, attributes, children}, and should either return a similar tuple or `nil` to delete the current node.
+
+  ## Examples
+
+      iex> html = {"div", [], ["hello"]}
+      iex> Floki.traverse_and_update(html, fn {"div", attrs, children} -> {"p", attrs, children} end)
+      {"p", [], ["hello"]}
+
+      iex> html = {"div", [], [{"span", [], ["hello"]}]}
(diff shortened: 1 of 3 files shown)


ericlathrop:
I had formatted with Elixir 1.8.1, and it looks like Travis wants formatting from an earlier version.

lib/floki.ex
```diff
@@ -228,6 +228,22 @@ defmodule Floki do
...
+      iex> html = {"div", [], ["hello"]}
+      iex> Floki.traverse_and_update(html, fn {"div", attrs, children} -> {"p", attrs, children} end)
+      {"p", [], ["hello"]}
+
+      iex> html = {"div", [], [{"span", [], ["hello"]}]}
+      iex> Floki.traverse_and_update(html, fn {"span", _attrs, _children} -> nil; tag -> tag end)
+      {"div", [], []}
+  """
+
+  defdelegate traverse_and_update(html_tree, fun), to: Floki.Traversal
```
philss: Nice!

lib/floki/traversal.ex
```diff
@@ -0,0 +1,23 @@
...
+  end
+
+  def traverse_and_update([], _fun), do: []
+
+  def traverse_and_update({elem, attrs, children}, fun) do
+    mapped_children = traverse_and_update(children, fun)
+    fun.({elem, attrs, mapped_children})
+  end
+
+  def traverse_and_update({:comment, children}, fun) do
```
philss: we can have some other cases like `{:pi, _, _}` and `{:pi, _}`. So I think you could have a clause similar to the first one, but in the bottom.

```elixir
def traverse_and_update(html_node, _fun), do: html_node
```

philss: I'm going to fix this after merge.

PR #220, opened 2019-09-11
Bring over mochiweb_html and remove dependencies

josevalim:
This commit brings the mochiweb_html module and remove
the dependencies on other modules by proper equivalents
in Erlang or Elixir. Every change on the original file
has been annotated with comments, other than that the
functionality remains the same.

### DIFF
--- a/src/floki_mochi_html.erl
+++ b/src/floki_mochi_html.erl
@@ -1,0 +1,829 @@
+%% @author Bob Ippolito <bob@mochimedia.com>
+%% @copyright 2007 Mochi Media, Inc.
+%%
+%% Permission is hereby granted, free of charge, to any person obtaining a
+%% copy of this software and associated documentation files (the "Software"),
+%% to deal in the Software without restriction, including without limitation
+%% the rights to use, copy, modify, merge, publish, distribute, sublicense,
+%% and/or sell copies of the Software, and to permit persons to whom the
+%% Software is furnished to do so, subject to the following conditions:
+%%
+%% The above copyright notice and this permission notice shall be included in
+%% all copies or substantial portions of the Software.
+%%
+%% THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+%% IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+%% FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+%% THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+%% LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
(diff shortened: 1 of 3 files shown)


josevalim:
Hi @philss! Thanks for the work on Floki!

We are evaluating Floki to be used on Phoenix LiveView and one of the concerns we had was the large mochiweb dependency. This PR extracts the dependency into a single file which contains a proper copy of the license. The file was also slightly changed to avoid dependencies on other modules from mochiweb and reduce the footprint.

It seems there is some work to remove mochiweb altogether but I hope you will consider this PR as an alternative meanwhile. If there are any bugs in the future as consequence of this replacement, feel free to ping me. :)

src/floki_mochi_html.erl
```diff
@@ -0,0 +1,829 @@
...
+tokenize_charref(Bin, S=#decoder{offset=O}) ->
+    try
+        case tokenize_charref_raw(Bin, S, O) of
+            {C1, S1} when C1 >= 16#D800 andalso C1 =< 16#DFFF ->
+                %% Surrogate pair
+                tokeninize_charref_surrogate_pair(Bin, S1, C1);
+            {Unichar, S1} when is_integer(Unichar) ->
+                %% CHANGED: Previously this was mochiutf8:codepoint_to_bytes(Unichar)
+                %% but that is equivalent to the below.
+                {{data, <<Unichar/utf8>>, false},
```
josevalim: CHANGED.

src/floki_mochi_html.erl
```diff
@@ -0,0 +1,829 @@
...
+                                         orelse C =:= $/
+                                         orelse C =:= $> ->
+            throw(invalid_charref);
+        <<_:O/binary, $;, _/binary>> ->
+            Len = O - Start,
+            %% CHANGED: Previously this was mochiweb_charref:charref/1
+            %% but the functionality below is equivalent;
+            <<_:Start/binary, Raw:Len/binary, _/binary>> = Bin,
+            <<CP/utf8>> = 'Elixir.HtmlEntities':decode(<<$&, Raw/binary, $;>>),
+            {CP, ?INC_COL(S)};
```
josevalim: CHANGED.

src/floki_mochi_html.erl
```diff
@@ -0,0 +1,829 @@
...
+    escape_attr(atom_to_list(A), []);
+escape_attr(S) when is_list(S) ->
+    escape_attr(S, []);
+escape_attr(I) when is_integer(I) ->
+    escape_attr(integer_to_list(I), []);
+%% CHANGE: Previously it used mochinum:digits/1 but float_to_list/1 is
+%% good enough generally. In case it isn't, a similar alternative would
+%% be lists:flatten(io_lib:format("~p", [F])).
+escape_attr(F) when is_float(F) ->
+    escape_attr(erlang:float_to_list(F), []).
```
josevalim: CHANGED.

philss:
@josevalim Thank you very much for this PR (and all the work in the community)! :purple_heart: 

I think it's a good move, because it avoid problems we had in the past (https://github.com/philss/floki/issues/68) and reduces the compile time for our users.

And also, it's great to know that you are considering using Floki on Phoenix LiveView! Are you trying to remove the front-end dependencies?

Please let me know if I can help in this work. Once again, thank you!

Issue #209, opened 2019-07-23
Floki.raw_html/2 doesn't correctly encode attributes with both single and double quotes


ericlathrop:
Example:

```elixir
iex(21)> "<span data-stuff=\"&quot;&#39;\">" |> Floki.parse |> Floki.raw_html  
"<span data-stuff='\"''></span>"
```

Expected either:
```elixir
"<span data-stuff='\"&#39;'></span>"
```
or
```elixir
"<span data-stuff=\"&quot;'\"></span>"
```

philss:
@ericlathrop sorry for the delay. Thank you for open it. I believe this can be fixed almost in the same way that was in [Meeseeks](https://github.com/mischov/meeseeks/pull/64).

danhuynhdev:
Hi @ericlathrop, I've created a PR (#227) for this. Hope to get some feedback from you.

philss:
@danhuynhdev Thank you! :purple_heart: 

I'm going to take a look this week, I hope. It looks like it fixes the problem, but I want to check some scenarios.

Issue #181, opened 2018-05-08
Loosing information about single/double quotes in the attribute value


grych:
When parsing, Floki loses the information if the attribute was introduced as single or double quoted text. This is generating an issue when rendering attributes with double quotes inside.

Observe the example:
```elixir
iex> html = "<tag attr='one \"two\" three'>"
"<tag attr='one \"two\" three'>"
iex> Floki.attribute(html, "attr")
["one \"two\" three"]
```
Attribute is retrieved correctly, as a proper Elixir string. But let's try to re-render the parsed HTML:
```elixir
iex> rerendered = html |> Floki.parse() |> Floki.raw_html()
"<tag attr=\"one \"two\" three\"></tag>"
```
This is a different html than it was in the beginning. And the attribute value has changed:
```elixir
iex> Floki.attribute(rerendered, "attr")
["one "]
```

This behaviour causes this issue: grych/drab#117. Notice that using quotes inside the attribute value is [correct](https://www.w3schools.com/html/html_attributes.asp).

I understand that Floki does not store the information about single/double quotes used in the attribute, but maybe the simple workaround exists, like **if there is a double quote inside the attribute value, render it with single quote**?

mischov:
Good point. I can't think of a better solution.

grych:
It should do the trick, as no other escaping single or double quotes is permitted in the attribute value, as well as the value containing both single and double quote in the same time is not allowed.

philss:
@grych good catch! I think your solution should work good. Please go ahead with a PR. Thank you!

grych:
@philss will do!

grych:
Fixed in #182. 
Please let me know when it comes into the release, as I need to update deps for Drab. Thanks!

PR #227, opened 2019-10-06
Escape problematic characters when encoding to html

danhuynhdev:
Fixed: #209 

With reference from https://github.com/mischov/meeseeks/pull/64

### DIFF
--- a/lib/floki/raw_html.ex
+++ b/lib/floki/raw_html.ex
@@ -35,8 +35,7 @@ defmodule Floki.RawHTML do
 
   defp build_raw_html([], html, _encoder), do: html
 
-  defp build_raw_html(string, _html, encoder) when is_binary(string),
-    do: html_escape_text(encoder.(string))
+  defp build_raw_html(string, _html, encoder) when is_binary(string), do: encoder.(string)
 
   defp build_raw_html(tuple, html, encoder) when is_tuple(tuple),
     do: build_raw_html([tuple], html, encoder)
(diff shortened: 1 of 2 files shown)


lib/floki/raw_html.ex
```diff
@@ -34,7 +34,8 @@ defmodule Floki.RawHTML do
   end
 
   defp build_raw_html([], html, _encoder), do: html
-  defp build_raw_html(string, _html, encoder) when is_binary(string), do: encoder.(string)
+  defp build_raw_html(string, _html, encoder) when is_binary(string),
+    do: html_escape_text(encoder.(string))
```
danhuynhdev: Not sure if this is needed as I see that spec already covers this.
https://github.com/philss/floki/blob/6700c14f842765eb5d1661af18fd686e87d1ddd7/test/floki_test.exs#L293

lib/floki/raw_html.ex
```diff
@@ -88,11 +89,7 @@ defmodule Floki.RawHTML do
   end
 
   defp build_attrs({attr, <<value::binary>>}, attrs) do
-    if String.contains?(value, "\"") do
-      ~s(#{attrs} #{attr}='#{value}')
-    else
-      ~s(#{attrs} #{attr}="#{value}")
-    end
+    "#{attrs} #{attr}=\"#{html_escape_attribute_value(value)}\""
```
danhuynhdev: Always wrap with `"` instead of alternate between the two. The original issue (#181) that introduces this `if` should also be covered by this fix.

philss:
@danhuynhdev awesome! Thank you! 💜

PR #230, opened 2019-10-10
Add github actions workflow

### LINKED ISSUE #228: Replace Travis CI with GitHub actions
GitHub Actions is a powerful tool, and seems to be more integrated with GitHub. I think it would be great to give a try and replace our Travis CI setup with a GitHub workflow.

I'm not sure if one can run the tests in her/his fork.

Docs about:
- [About GitHub Actions](https://help.github.com/en/articles/about-github-actions)
- [The Elixir action](https://github.com/actions/setup-elixir) - the Matrix example on README is similar to what we want.

donaldducky:
fixes #228

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,39 @@
+name: CI
+
+on: [push, pull_request]
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+
+    container:
+      image: elixir:${{ matrix.elixir }}-slim
+
+    name: Elixir ${{ matrix.elixir }}
+
+    strategy:
+      matrix:
+        elixir: [1.9.1, 1.8.2, 1.7.4, 1.6.6, 1.5.3]
+
+    steps:
+      - uses: actions/checkout@v1.0.0
+
+      - name: Install dependencies
+        run: |-
+          mix local.rebar --force
+          mix local.hex --force
+          mix deps.get
+
+      - name: Check format
+        if: matrix.elixir >= 1.6
+        run: mix format --check-formatted
+
+      - name: Run tests
+        run: |-
+          mix test
+
+      - name: Run inch.report
+        if: matrix.elixir >= 1.7
+        run: |-
+          mix deps.get --only docs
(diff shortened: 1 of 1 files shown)


donaldducky:
Hey @philss 

I tried using https://github.com/actions/setup-elixir but it doesn't seem quite ready yet.

When trying to use the matrix in the travis-ci config, it was failing to install some of the OTP versions.

I'm not sure if this workflow is enough for switching over to workflows yet since it doesn't test multiple OTP versions with each elixir version in the matrix but this is a start.

I ended up using the official elixir docker containers for the environment in the workflow.

Some other notes:
* added elixir 1.9.1 and 1.8.2 to the list (maybe can go with minor releases instead of patch releases?)
* a couple steps have `if` conditionals checking against the elixir version to decide whether or not to run `mix format` (requires >= 1.6) and `inch.report`(requires >= 1.7)

Let me know if you have any thoughts or ideas on this!

donaldducky:
Hmm, I had to change the elixir versions to be minor versions for the `if` comparisons.

Github converts the values to numbers for comparisons "if the types don't match".
https://help.github.com/en/articles/contexts-and-expression-syntax-for-github-actions#operators

It can't convert 1.9.1 to a number based on:

> Parsed from any legal JSON number format, otherwise NaN. 
Note: empty string returns 0.

I also noticed there's no "Actions" tab in this repo 🤔

donaldducky:
Not sure why that last build failed (maybe some kind of transient network error?).

Here's what the github actions look like for me:
Screenshot 2019-10-09 22 06 23

philss:
@donaldducky I think it's good enough to start! Maybe we don't need to test against the OTP versions after the 20.x since they are very stable.

Thank you! 💜 

(PS: I'm going to merge but I don't now if the repo will have Actions activated, since it's still in beta).

donaldducky:
No problem 👍 !

I think you have to have registered for the beta and got accepted for it to work (that's how I have them enabled).

philss:
@donaldducky Thank you! It works now. 🚀

Issue #237, opened 2019-11-18
Floki mistakes &adxnnl=1; in attributes for character code and crashes


umuro:
## Description
attribute contents crash Floki when they contain & and later ;.  For example "&adxnnl=1;" .The tag example below is from https//:nytimes.com
## To Reproduce

- Using Floki v0.23.0
- Using Elixir v1.9.1
- Using Erlang OTP v22.1.1
- With this code:
  ```
"<meta data-rh=\"true\" name=\"msapplication-task\" content=\"name=Homepage;action-uri=https://www.nytimes.com?src=iepin&adxnnl=1;icon-uri=https://static01.nyt.com/images/icons/homepage.ico\"/>" |> Floki.parse
  ```

## Expected behavior

Floki should not mistake it for a special character whenever there is an & around... In some sites, query strings are used as attribute contents. Query strings mean a lot of &'s around.

## Workaround
Before Floki.parse get rid of the annoying pattern
```
 ~r/&(?=[[:alnum:]]+=.+;)/ |> Regex.replace(string, "\+\*\+\*")
```
It's easy to revert this also. But a lot of precious CPU time is lost

philss:
@umuro Thank you for opening the issue! It was fixed in version 0.23.1. Can you try again with that version?

Issue #235, opened 2019-11-08
Floki fails to parse Some HTML


Adzz:
## Description

If I try to run `Floki.parse` on this html it fails:

```html
<!DOCTYPE html>
<html lang="en-GB" class="no-js ">
  <head> </head>
  <body>
    <br />*T&C&#39;s Apply.<br />
  </body>
</html>
```

I get the error:

```sh
** (MatchError) no match of right hand side value: "&C&#39;"
    (floki) src/floki_mochi_html.erl:705: :floki_mochi_html.tokenize_charref_raw/3
    (floki) src/floki_mochi_html.erl:651: :floki_mochi_html.tokenize_charref/2
    (floki) src/floki_mochi_html.erl:306: :floki_mochi_html.tokens/3
    (floki) src/floki_mochi_html.erl:83: :floki_mochi_html.parse/1
    (floki) lib/floki/html_parser/mochiweb.ex:7: Floki.HTMLParser.Mochiweb.parse/1
```

- Using Floki 0.23.0
- Using Elixir 1.9.0
- Using Erlang OTP 22.0.5

## Expected behavior

I think it's valid HTML, so it should parse

Adzz:
Digging further it looks like this is the line that is failing:

```erlang
<<CP/utf8>> = 'Elixir.HtmlEntities':decode(<<$&, Raw/binary, $;>>),
```
in `floki/src/floki_mochi_html.erl`

I don't know erlang, but does this mean that it wont parse non UTF8 strings?

oskar1233:
Just to confirm, I have `MatchError` on the same line (`src/floki_mochi_html.erl:705`) with `"&dash;` right hand side value.

Might be somehow related that my HTML had `\uFEFF` character on 1st position but the error occurs even after stripping it out.

Adzz:
Hello, from more investigation it looks like we can strip the failing cases to being html that has a `&` in it. 

It looks like when we hit a `&` the parser assumes that are about to hit one of these things `&#39;` (I'm not sure what the name is? An encoded character??). HOWEVER We can sometimes legitimately be hitting a `&` followed by an escaped apostrophe, for example if in the HTML we say something like `Check out these T&C's` (terms and conditions). In that case the apostrophe would be escaped to this `&#39;` meaning Floki sees it as `T&C&#39;s`. It then incorrectly sees the first `&` and thinks "this must be escaped html". 

I have no idea how to fix.

oskar1233:
@Adzz 
[...]

PR #239, opened 2019-12-01
Fix parsing when cannot decode a charref correctly

### LINKED ISSUE #233: Issue with HTML entity in link URL query parameter in floki 0.23.0
## Description

It seems like (as far as I can tell), having an HTML entity in a URL's query parameter value for a link breaks the parser starting with floki 0.23.0.  This might also happen in other circumstances.  This might not be valid HTML, but I did run into it in the wild and it's probably not great that the parser crashes, especially when it worked previously.

## To Reproduce

Using:

```
Erlang/OTP 21 [erts-10.3.5] [source] [64-bit] [smp:12:12] [ds:12:12:10] [async-threads:1] [hipe] [dtrace]

Elixir 1.8.1 (compiled with Erlang/OTP 20)
```

When I try to parse this HTML:

```html
<a href="http://foo.com/blah?hi=blah&foo=&#43;Park" class="foo">test</a>
```

... in floki 0.23.0 I get this error:

```
** (MatchError) no match of right hand side value: "&foo=&#43;"
    src/floki_mochi_html.erl:705: :floki_mochi_html.tokenize_charref_raw/3
    src/floki_mochi_html.erl:651: :floki_mochi_html.tokenize_charref/2
    src/floki_mochi_html.erl:546: :floki_mochi_html.tokenize_quoted_attr_value/4
    src/floki_mochi_html.erl:515: :floki_mochi_html.tokenize_attributes/3
    src/floki_mochi_html.erl:362: :floki_mochi_html.tokenize/2
    src/floki_mochi_html.erl:306: :floki_mochi_html.tokens/3
    src/floki_mochi_html.erl:83: :floki_mochi_html.parse/1
    lib/floki/html_parser/mochiweb.ex:7: Floki.HTMLParser.Mochiweb.parse/1
```

## Expected behavior

With 0.22.0 I get the parsed result as expected:

```elixir
{"a", [{"href", "http://foo.com/blah?hi=blah&foo=+Park"}, {"class", "foo"}],
 ["test"]}
```

## Standard Thanks

Thanks so much for floki!

philss:
It throws an error to the parser so it can ignore the decode of charref
when it can't get back a valid UTF8 codepoint.

This fixes #233, #235 and #237.

### DIFF
--- a/src/floki_mochi_html.erl
+++ b/src/floki_mochi_html.erl
@@ -702,8 +702,13 @@ tokenize_charref_raw(Bin, S=#decoder{offset=O}, Start) ->
             %% CHANGED: Previously this was mochiweb_charref:charref/1
             %% but the functionality below is equivalent;
             <<_:Start/binary, Raw:Len/binary, _/binary>> = Bin,
-            <<CP/utf8>> = 'Elixir.HtmlEntities':decode(<<$&, Raw/binary, $;>>),
-            {CP, ?INC_COL(S)};
+
+            case 'Elixir.HtmlEntities':decode(<<$&, Raw/binary, $;>>) of
+              <<CP/utf8>> ->
+                {CP, ?INC_COL(S)};
+              _ ->
+                throw(invalid_charref)
+            end;
         _ ->
             tokenize_charref_raw(Bin, ?INC_COL(S), Start)
     end.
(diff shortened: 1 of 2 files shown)


minibikini:
@philss, please release

philss:
@minibikini please check the version 0.23.1. It was released with this fix. 🚀

minibikini:
Thanks a lot, @philss! I can confirm the new version has fixed [our problem](https://git.pleroma.social/pleroma/pleroma/issues/1451).

Adzz:
Thanks so much! ❤️

Issue #236, opened 2019-11-08
Floki.parse differs when using html5ever


andyleclair:
## Description

Mochiweb Floki will produce different output than html5ever, namely, the output of `Floki.parse` will be wrapped in `<html><head></head><body>...</body></html>`

## To Reproduce


- Using Floki v0.23.0
- Using html5ever
- Using Elixir v1.9.3
- Using Erlang OTP v21.3.8.9
- With this code:

```elixir
defmodule TestCases do
  @test_cases [
    {
      ~s[<a href="javascript:alert('XSS');">Click here</a>],
      ~s[<a href="#">Click here</a>]
    },
    {
      ~s[<a href="whatever" onclick="alert('XSS');">Click here</a>],
      ~s[<a href="whatever">Click here</a>],
    },
    {
      ~s[<body onload="alert('XSS')"><p>Hello</p></body>],
      ~s[<body><p>Hello</p></body>],
    },
    {
      ~s[<img src="javascript:alert('XSS');">],
      ~s[<img src="#"/>],
    },
    {
      ~s[<script>alert('XSS');</script>],
      ~s[],
    },
    {
      ~s[<body background="javascript:alert('XSS');"><p>Hello</p></body>],
      ~s[<body background="#"><p>Hello</p></body>],
    },
    {
      ~s[<style>body { background-image: expression('alert("XSS")'); }</style>],
      ~s[<style>body { background-image: removed_by_strip_js('alert("XSS")'); }</style>],
    },
    {
      ~s[<style>body { background-image: url('javascript:alert("XSS")'); }</style>],
      ~s[<style>body { background-image: url('removed_by_strip_js:alert("XSS")'); }</style>],
    },
    {
      ~s[<style><script>alert('XSS')</script></style>],
      ~s[<style><script>alert('XSS')</script></style>],
    },
    {
      ~s[<style> h1 > a { color: red; } </style>],
      ~s[<style> h1 > a { color: red; } </style>],
    },
    {
      ~s[<],
      ~s[&lt;],
    },
    {
      ~s[>],
      ~s[&gt;],
    },
    {
      ~s[],
      ~s[],
    },
  ]

  def test_cases, do: @test_cases
end

TestCases.test_cases |> Enum.map(fn {ins, _outs} -> Floki.parse(ins) end)

[                                                                                                                                                                                                                                                                                         
  [                                                                                                                                                                                                                                                                                       
    {"html", [],                                                                                                                                                                                                                                                                          
[...]

PR #240, opened 2019-12-02
Add fast_html parser and test all parsers on CI

### LINKED ISSUE #156: Test with html5ever HTML parser on CI

### LINKED ISSUE #234: Support fast_html as an optional html parser
## Feature goal

[`fast_html`](https://git.pleroma.social/pleroma/fast_html) is a C-Node binding for lexborisov's [myhtml](https://github.com/lexborisov/myhtml), it's from 1.2 to 2 times faster than html5ever ([benchmark link](https://hexdocs.pm/fast_html/readme.html#benchmarks)) and requires only a c complier and gmake. 

I can open a PR if you have no objection to adding it.

## Dependencies

```elixir
{:fast_html, "~> 0.99"}
```

:)

rinpatch:
Closes #234, closes #156 

Note that:
1. Currently html5ever and fast_html are failing exactly the same tests, I am guessing that's because of #236 and I think this should be solved in a separate PR.
2. The test code is pretty hacky, would be nice if `System.cmd` could be avoided, but I can't figure out why is `ExUnit` not running the tests without it.

### DIFF
--- a/lib/floki/html_parser/fast_html.ex
+++ b/lib/floki/html_parser/fast_html.ex
@@ -1,0 +1,16 @@
+defmodule Floki.HTMLParser.FastHtml do
+  @moduledoc false
+
+  def parse(html) do
+    case Code.ensure_loaded(:fast_html) do
+      {:module, module} ->
+        case module.decode(html) do
+          {:ok, result} -> result
+          {:error, _message} = error -> error
+        end
+
+      {:error, _reason} ->
+        raise "Expected module :fast_html to be available."
+    end
+  end
+end
--- a/mix.exs
+++ b/mix.exs
@@ -13,6 +13,7 @@ defmodule Floki.Mixfile do
       elixir: "~> 1.5",
       package: package(),
       deps: deps(),
+      aliases: aliases(),
       source_url: "https://github.com/philss/floki",
       docs: [extras: ["README.md"], main: "Floki", assets: "assets"]
     ]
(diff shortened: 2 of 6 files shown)


rinpatch:
Oh, I didn't notice you have both github workflows and travis. Why is that? Should I also modify the workflow?

rinpatch:
Hm, not sure what's wrong with github actions now, they just don't trigger for some reason.

rinpatch:
Apologies for the commit spam, haven't worked with Github Actions before

.travis.yml
```diff
@@ -9,6 +9,10 @@ otp_release:
   - 20.3
   - 20.0
   - 19.3
+env:
```
philss: I forgot to remove Travis CI from the project after adding GitHub Actions 😬 
I think this configuration is not needed.

lib/floki/html_parser/fast_html.ex
```diff
@@ -0,0 +1,16 @@
+defmodule Floki.HTMLParser.FastHtml do
+  @moduledoc false
+
+  def parse(html) do
```
philss: I'm about to change those interface modules to have a `parse_document/2` and `parse_fragment/2`, and remove the `parse/1`. Not sure when, but this should be made in order to fix the different output from parsers in test.

mix.exs
```diff
@@ -23,13 +24,60 @@ defmodule Floki.Mixfile do
   end
 
   defp deps do
+    # Needed to avoid installing unnecessary deps on travis
```
philss: If we don't need Travis, I think we can remove this too, right?

mix.exs
```diff
@@ -23,13 +24,60 @@ defmodule Floki.Mixfile do
...
+          |> Macro.underscore()
+
+        {{:"test.#{cli_name}", &test_with_parser(parser, &1)}, [cli_name | acc]}
+      end)
+
+    aliases
+    |> Keyword.put(:test, &test_with_parser(cli_names, &1))
+  end
+
+  # Hardcoded because we can't load the floki application and get the module list at this point.
```
philss: Can you declare this list at line 45 directly?

mix.exs
```diff
@@ -23,13 +24,60 @@ defmodule Floki.Mixfile do
...
+  defp get_parsers() do
+    [Floki.HTMLParser.Mochiweb, Floki.HTMLParser.FastHtml, Floki.HTMLParser.Html5ever]
+  end
+
+  # Hack: If Mix.Task.Test.run is called the second time, it won't run any tests
```
philss: This is probably related to the `ExUnit` that is started at `test/test_helper.exs` 🤔 
It's OK to leave without the line comment here.

mix.exs
```diff
@@ -23,13 +24,60 @@ defmodule Floki.Mixfile do
...
+  defp test_with_parser(parser_cli_names, args) when is_list(parser_cli_names) do
+    Enum.each(parser_cli_names, fn cli_name ->
+      Mix.shell().cmd("mix test.#{cli_name} --color #{Enum.join(args, " ")}",
+        env: [{"MIX_ENV", "test"}]
+      )
+    end)
+  end
+
+  defp test_with_parser(parser, args) do
+    Mix.shell().info("Running tests with #{parser}")
```
philss: Nice! 👍

test/test_helper.exs
```diff
@@ -1 +1,7 @@
+# fast_html uses a C-Node worker for parsing, so starting the application
```
philss: 👍

.travis.yml
rinpatch: Ah, should I just remove `.travis.yml`?

lib/floki/html_parser/fast_html.ex
rinpatch: Yeah I saw your comment. `fast_html` doesn't support parsing fragments now, but it's on my list to add right after this is merged.

mix.exs
rinpatch: Applies to Gitlab Actions as well, I just wrote this comment when I didn't know you had it :)

lib/floki/html_parser/fast_html.ex
rinpatch: Added fragment support in https://git.pleroma.social/pleroma/fast_html/merge_requests/14. Want me to add new functions now?

Also what is the second argument in `parse_document/2` supposed to be? I assume it's the context tag for `parse_fragment/2`, but you don't need that when parsing documents?

rinpatch:
@philss ping

philss:
@rinpatch pong. Sorry for the delay. My last two weeks are being crazy at work :/
I'm gonna look and respond better tomorrow.

lib/floki/html_parser/fast_html.ex
philss: I don't think we should do this in this pull request. It will be awesome to fix this in another PR, but I think we need also to support fragments for html5ever, which is not implemented. I will take a look and check if it's OK to do that in html5ever_elixir.
But let's think better on those functions in another context. And thank you for adding that feature!

rinpatch:
@philss rebased and added a readme entry

PR #242, opened 2020-01-08
Fix typo of FastHtml in the readme

fatboypunk:
Since `Floki.HTMLParser.FastHTML` doesn't exist


philss:
@fatboypunk thanks! 💜

PR #243, opened 2020-01-08
Add parse_document!/1 and parse_fragment!/1

schneiderderek:
Both functions have the same functionality as their respective
counterparts, but raise a `Floki.ParseError` in case of a parsing error
instead of returning an error tuple or return the successfully parsed
html tree.

This is especially useful when using the pipe operator.

### DIFF
--- a/lib/floki.ex
+++ b/lib/floki.ex
@@ -105,6 +105,28 @@ defmodule Floki do
   defdelegate parse_document(document), to: Floki.HTMLParser
 
   @doc """
+  Parses a HTML Document from a string.
+
+  Similar to `Floki.parse_document/1`, but raises `Floki.ParseError` if there was an
+  error parsing the document.
+
+  ## Example
+
+      iex> Floki.parse_document!("<html><head></head><body>hello</body></html>")
+      [{"html", [], [{"head", [], []}, {"body", [], ["hello"]}]}]
+
+  """
+
+  @spec parse_document!(binary()) :: html_tree()
+
+  def parse_document!(document) do
+    case parse_document(document) do
+      {:ok, parsed_document} -> parsed_document
+      {:error, message} -> raise Floki.ParseError, message: message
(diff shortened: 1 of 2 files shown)


lib/floki.ex
```diff
@@ -104,6 +104,28 @@ defmodule Floki do
 
   defdelegate parse_document(document), to: Floki.HTMLParser
 
+  @doc """
+  Parses a HTML Document from a string.
+
+  Similar to `c:parse_document/1`, but raises `Floki.ParseError` if there was an
```
philss: Why using `c:` instead of `Floki.`?

lib/floki.ex
```diff
@@ -113,7 +135,23 @@ defmodule Floki do
 
   @spec parse_fragment(binary()) :: {:ok, html_tree()} | {:error, String.t()}
 
-  defdelegate parse_fragment(document), to: Floki.HTMLParser
+  defdelegate parse_fragment(fragment), to: Floki.HTMLParser
+
+  @doc """
+  Parses a HTML fragment from a string.
+
+  Similar to `c:parse_fragment/1`, but raises `Floki.ParseError` if there was an
```
philss: Same question here 🤔

lib/floki.ex
```diff
@@ -113,7 +135,23 @@ defmodule Floki do
 
   @spec parse_fragment(binary()) :: {:ok, html_tree()} | {:error, String.t()}
 
-  defdelegate parse_fragment(document), to: Floki.HTMLParser
+  defdelegate parse_fragment(fragment), to: Floki.HTMLParser
```
philss: 👍

lib/floki.ex
schneiderderek: I mixed up the prefix for documentation in this case. I went ahead and fixed it, thanks! 😄

PR #254, opened 2020-01-31
Fix improper "it's" usage.

RichMorin:
The possessive "its", like "his" and "hers", requires no apostrophe.


philss:
@RichMorin Thank you! 💜

PR #268, opened 2020-04-11
improve language in contribution document

Devon-Olivier:


philss:
@Devon-Olivier thank you! 💜

PR #270, opened 2020-04-17
Support Encoding More Entities in Floki.raw_html

ntenczar:
`Floki.raw_html` appears to encode HTML entities on attributes using
`html_escape_attribute_value/1`, but the regular expression left out
`<`, `>`, and `'` despite having clauses that match those characters in
`html_escape_char/1`. This adds those three characters to the regular
expression match, and adds unit tests to that effect.

It's also possible this was intentional and I'm missing something, so if
so I'm happy to hear why and I can close this PR.

### DIFF
--- a/lib/floki/raw_html.ex
+++ b/lib/floki/raw_html.ex
@@ -117,7 +117,7 @@ defmodule Floki.RawHTML do
   # html_escape
 
   def html_escape_attribute_value(attribute_value) do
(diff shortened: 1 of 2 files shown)


philss:
@ntenczar great! Thank you! :purple_heart:

ntenczar:
awesome thanks for merging!

PR #273, opened 2020-04-30
if: expression on ci file configuration doesn't work as expected

Cleidiano:
The current configuration skip configuration on all steps.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -40,7 +40,7 @@ jobs:
         if: matrix.parser == 'html5ever'
 
       - name: Check format
-        if: matrix.elixir >= 1.8
+        if: (!startsWith(matrix.elixir, '1.6') && !startsWith(matrix.elixir, '1.7'))
         run: mix format --check-formatted
 
       - name: Run tests


Cleidiano:
I will check later why the test have been failed.

Cleidiano:
The issue were fixed on InchEx https://github.com/rrrene/inch_ex/pull/68, but not release yet.

Cleidiano:
@philss The fix for inch_ex was rollback, after it released `2.1` I can open up a new PR to fix it.

philss:
@Cleidiano awesome! Thank you! 💜 

I didn't get the problem with InchEx 🤔 
But feel free to open a new PR to fix it.

Cleidiano:
The `inch_ex` task on CI isn't running because the same problem I have fixed in `check-formatted` task, but current inch_ex have a bug that crash on parse module that haven't docs chunk on beam file, for example `floki_mochi_html.erl`.
To fix the inch_ex task we need to await for release `2.1` that fix de problem, for reference https://github.com/rrrene/inch_ex/pull/68

Issue #282, opened 2020-06-14
`Floki.filter_out` for text nodes


ckruse:
## Feature goal

It would be nice to be able to filter out text nodes. `:text` (in comparison to `:comment`) could be used.

```elixir
Floki.filter_out(doc, :text)
```

philss:
I totally forgot to close the issue. It was introduced by #283 . Thanks!

ckruse:
Me too. Thanks! :)

PR #283, opened 2020-06-14
feat: `Floki.filter_out/2` now can filter out nodes

ckruse:
Regarding #282

### DIFF
--- a/lib/floki/filter_out.ex
+++ b/lib/floki/filter_out.ex
@@ -49,7 +49,7 @@ defmodule Floki.FilterOut do
 
   defp filter({nodetext, _, _}, selector) when nodetext === selector, do: false
   defp filter({nodetext, _}, selector) when nodetext === selector, do: false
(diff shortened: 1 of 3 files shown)


lib/floki/filter_out.ex
```diff
@@ -49,6 +49,7 @@ defmodule Floki.FilterOut do
 
   defp filter({nodetext, _, _}, selector) when nodetext === selector, do: false
   defp filter({nodetext, _}, selector) when nodetext === selector, do: false
+  defp filter(text, :text) when is_bitstring(text), do: false
```
philss, line 52: Why not use `is_binary/1` here? Since we are always dealing with `utf8` in text, I think we could consider text as at least a binary. What do you think?

ckruse: You are right, it should be `is_binary` - I will push a fix.

PR #287, opened 2020-06-17
Remove duplication in traverse and update

josevalim:


philss:
@josevalim Thank you again! :purple_heart:

PR #299, opened 2020-08-14
Update traverse_and_update/2 and /3 docs and tests

hubertlepicki:
Since `traverse_and_update/2` and `traverse_and_update/3` can encounter
and process not just HTML tags, but also - for example HTML comments -
the function passed as argument has to be aware of these possibilities.
Otherwise the function will crash.

This code change updates documentation to make user aware that this can
and will happen when parsing HTML, and also updates examples/doctests to
always have the second function clause to handle arguments which are not
HTML tags.


philss:
@hubertlepicki awesome! Thank you! :purple_heart:

Issue #286, opened 2020-06-17
Revisit Floki.map API


josevalim:
The new`traverse_and_update` has been quite helpful, thank you, we are making extensive use of it in LiveView to optimize some traversals!

Given its introduction, it may be worth even visiting other parts of the API because I believe `Floki.map` can be completely replaced by `Floki.traverse_and_update`:

```elixir
  def map(html_tree_list, fun) do
    traverse_and_update(html_tree_list, fn
      {tag, attrs, children} ->
        {tag, attrs} = fun.({tag, attrs})
        {tag, attrs, children}

      other ->
        other
    end)
  end
```

The difference is that `traverse_and_update` is quite more efficient as it doesn't need to build the HTML tree. Given `map` has a shorter name, people may be more inclined towards using it, while using `traverse_and_update` would be wiser.

Another possible approach to take here is to deprecate `map` in favor of a `find_and_update(tree, selector, fun)` function, which works like `map` but updates only the nodes given by the selector, keeping the others intact. This would leverage the power of the HTML Tree and introduce a feature set not covered today. In fact, if there is a `find_and_update`, one could easily implement something like `Floki.attr/4` on top of `find_and_update`.

josevalim:
Oh, another cool idea. If `find_and_update` allows `fun` to return `nil` for removals (like traverse_and_update), then `filter_out` can be implemented as: `find_and_update(node, select, fn _ -> nil end)`. So it can really be a function used as building block for a bunch of other functionality. :)

philss:
@josevalim thank you for the issue and ideas! :purple_heart: 
Replacing `Floki.map/2` with a `Floki.find_and_update/3` seems the better way to go. I don't like very much the `Floki.map/2` function and I wanted to deprecate it before.

We can do this in some steps I think:
- change the implementation of `map/2` to use `traverse_and_update/2`;
- add the `find_and_update/3` and add the deprecation;
- after some versions we remove the `map/2` function.

What do you think?

josevalim:
If the plan is to remove `map`, i wouldn't change it to use `traverse_and_update/2`, because what if it introduces minor incompatibilities? For example, `traverse_and_update/2` and `find` traverse in different order (find does parent first, traverse children first). While that should not be a problem for `map`, you never know what people are doing, like using `map` for side-effects.

philss:
It makes sense. So let's go for the deprecation/removal without the refactoring o/

philss:
[...]

PR #308, opened 2020-10-02
Deprecate Floki.map/2 and introduce Floki.find_and_update/3

philss:
This PR deprecates the `Floki.map/2` function in in favour of `Floki.find_and_update/3` or `Enum.map/2`.
This follows the ideas of #286 

With `find_and_update/3` we can refactor functions like `Floki.filter_out/2` to use it.

### DIFF
--- a/lib/floki.ex
+++ b/lib/floki.ex
@@ -333,7 +333,7 @@ defmodule Floki do
       iex>   other ->
       iex>     other
       iex> end)
-      iex> # => [{"a", [{"href", "https://elixir-lang.com"}], ["Elixir"]}]
+      [{"a", [{"href", "https://elixir-lang.com"}], ["Elixir"]}]
   """
 
   @spec find_and_update(
--- a/lib/floki.ex
+++ b/lib/floki.ex
@@ -228,7 +228,7 @@ defmodule Floki do
 
   @doc """
   Changes the attribute values of the elements matched by `selector`
-  with the function `mutation` and returns the whole element tree
+  with the function `mutation` and returns the whole element tree.
 
   ## Examples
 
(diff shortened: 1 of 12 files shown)


lib/floki.ex
```diff
@@ -304,25 +304,68 @@ defmodule Floki do
...
+      Enum.map(results, fn
+        html_node = %Floki.HTMLTree.HTMLNode{} ->
+          case fun.({html_node.type, html_node.attributes}) do
+            {updated_tag, updated_attrs} ->
+              {:update, %{html_node | type: updated_tag, attributes: updated_attrs}}
+
+            nil ->
+              {:delete, html_node}
+
+            _ ->
```
josevalim, line 357: Are you sure you want to ignore invalid results? :D

lib/floki.ex
```diff
@@ -304,25 +304,68 @@ defmodule Floki do
...
+    {tree, results} = Finder.find(html_tree, selector)
+
+    operations_with_nodes =
+      Enum.map(results, fn
+        html_node = %Floki.HTMLTree.HTMLNode{} ->
+          case fun.({html_node.type, html_node.attributes}) do
+            {updated_tag, updated_attrs} ->
+              {:update, %{html_node | type: updated_tag, attributes: updated_attrs}}
+
+            nil ->
```
josevalim, line 354: Maybe returning `:delete` is clearer than `nil`? 🤔

josevalim:
This looks good, just two minor comments. I also think it may be already worth to already change filter_out and friends to use the new function, to validate the function actually behaves like you expect.

lib/floki.ex
philss: I'm not sure /o\
Maybe letting it crash for invalid match? :thinking:

lib/floki.ex
philss: yeah, I was unsure if I should make it different from "traverse_and_update". Gonna do that and possible change "traverse_and_update" in the near future.

lib/floki.ex
josevalim: :+1: for let it crash!

philss:
@josevalim thanks! I changed `filter_out` and `attr` to use the new function, but I left a special case of `filter_out` (the comment/text one) to another refactor since it would be more complex for now.

josevalim:
Ship it!

philss:
I forgot about reorganizing the commits :disappointed:

PR #315, opened 2020-11-04
Remove unused guards on CI configuration

Cleidiano:
The guard `matrix.elixir >= 1.7` prevented the `inch.report` task to run on all pipelines, since elixir `1.7` isn't list in the matrix I removed the guard and then the task started run.
I upgraded inch_ex because of known bug on version 2.0.0 when the beam file doesn't have the doc chunk and fixed at version 2.1.0-rc.1.


philss:
@Cleidiano good call! Thank you! :purple_heart:

PR #317, opened 2020-11-19
start gleam wrapper

CrowdHailer:
I'm using floki in a Gleam project. Wondered if these wrappers might be useful. Currently I only need the query API's so that is what I have started with.

I don't think there is any downside from the standpoint of an Elixir application supporting these wrappers, the gleam files should be ignored by default.


philss:
@CrowdHailer awesome! Do we need to include this file in the Hex release as well? :smiley:

CrowdHailer:
I think, yes we do.

Will try and iron the kinks out over the next few days

mix.exs
```diff
@@ -12,6 +12,8 @@ defmodule Floki.Mixfile do
       description: @description,
       elixir: "~> 1.8",
       package: package(),
+      erlc_paths: ["src", "gen"],
+      compilers: [:gleam | Mix.compilers()],
```
CrowdHailer, line 16: This PR now works in my Gleam project, however these changes break it in an Elixir project. perhaps not surprisingly.

```
** (Mix) The task "compile.gleam" could not be found. Did you mean "compile.erlang"?
```

I think the solution is to only add the gleam compiler option if it is required by the calling library.

The options I have thought of are.
- An env variable 
- add the compiler if the module `Mix.Gleam` is defined, because using https://github.com/gleam-lang/mix_gleam is how gleam is added to mix projects. 
- some other option in mix dependencies I haven't though of.

mix.exs
```diff
@@ -43,6 +45,7 @@ defmodule Floki.Mixfile do
 
     [
       {:html_entities, "~> 0.5.0"},
+      {:mix_gleam, "~> 0.1.0"},
```
CrowdHailer, line 48: Testing this as a solution to always compile Gleam source code, though the best solution is probably to compile gleam code conditionally.

CrowdHailer:
It was simpler than I thought, the only concession to adding gleam to a project is to indicate that some erlang source files are in the `gen` directory. This is where they end up after being compiled from `src/*.gleam` files.

philss:
@CrowdHailer awesome! I think we can go with it. :rocket: 
I have two small issues that we may want to solve: 
- we need to add the file under `gen` directory to the list of files in the `package/0` function at `mix.exs`;
- would be nice to have a test case that compiles a small gleam project in order to ensure the compilation is working.

This second item is not so important for now, I think.

philss:
> we need to add the file under gen directory to the list of files in the package/0 function at mix.exs;

I'm sorry. This does not makes sense. I thought the file was under `gen`.

Issue #249, opened 2020-01-25
way to generate prettified HTML from a parse tree


RichMorin:
## Feature goal

I'd like something that works like `raw_html/1`, but does a reasonable job of formatting the result so that I can read it more easily.  I don't care whether this is implemented as an option to the current function or a new function.

-r

PR #323, opened 2020-12-21
Adds pretty option to raw_html

evaldobratti:
This is an attempt to solve https://github.com/philss/floki/issues/249

I think there is still room to improve, like making the padding character and line ending customizable, but I'd like to know if this is what you guys expect from this issue and if the way I tried to solve it is acceptable.

### DIFF
--- a/test/floki_test.exs
+++ b/test/floki_test.exs
@@ -298,6 +298,64 @@ defmodule FlokiTest do
     assert Floki.raw_html(tree, encode: false) == input
   end
 
+  test "raw_html pretty with doctype" do
+    html = """
+      <!doctype html>
+      <html>
+      <head>
+      <title>Test</title>
+      </head>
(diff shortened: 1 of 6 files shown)


test/floki_test.exs
```diff
@@ -298,6 +298,64 @@ defmodule FlokiTest do
...
+      </body>
+      </html>
+    """
+
+    pretty_html =
+      html
+      |> document!()
+      |> Floki.raw_html(pretty: true)
+
+    assert pretty_html == """
```
philss, line 334: In the future I think we can add the doctype to generated documents. I think the problem today is that we don't keep it when parsing.

philss (approved):
@evaldobratti awesome! Thank you! :D :purple_heart: 
I think this is a good start! I will merge and we can discuss improvements later.

Issue #322, opened 2020-12-17
Possible wrong typespec for `attribute/2`


mtarnovan:
The typespec of `attribute/2` is  ` @spec attribute(binary | html_tree, binary) :: list`, but the function can also be called with type `html_node` as first arg:

```elixir
iex(10)> Floki.attribute([{"a", [{"href", "https://google.com"}], ["Google"]}], "a", "href")
["https://google.com"]
iex(11)> Floki.attribute({"a", [{"href", "https://google.com"}], ["Google"]}, "a", "href")  
["https://google.com"]
```

Is the correct typespec actually ` @spec attribute(binary | html_tree | html_node, binary) :: list` or am I missing something? `traverse_and_update` for example, actually passes in a `html_node`, not a `html_tree` to the reducer, so doing this:
```
    Floki.traverse_and_update(doc, fn
      {"a", _attrs, _children} = el ->
        Floki.attribute(el, "src")
```
would actually trigger an error from dialyxir.

edit: this may apply to other funs as well

philss:
@mtarnovan you are correct! :+1: Although we are going to remove the support for binaries, this function accepts a `html_node()`. Another thing to mention is that `Floki.find/2` also accepts a `html_node()`, so I think we can fix that too. Do you mind to open a PR? Thanks!

mtarnovan:
Sure, I can open a PR. Just fix the typespecs or did you have something more in mind?

philss:
@mtarnovan Just the typespecs for now. Thank you! :purple_heart:

PR #326, opened 2021-01-04
Fix typespecs for `find`, `attr` and `attribute` to also allow html_n…

mtarnovan:
…ode #322


philss:
@mtarnovan thank you! :purple_heart:

PR #342, opened 2021-04-02
Mochiweb - parse style tags as plaintext

### LINKED ISSUE #339: HTML tags inside <style> elements get interpreted as children.
## Description

I was not writing any Floki code my self, but actually using [Premailex](https://github.com/danschultzer/premailex).
Premailex was throwing an error because it was not ready to find any children inside style tags other than strings, after some debugging I found out Floki was looking at an html tag in my css and parsed that as a children.

## To Reproduce

- Using Floki v0.30.0
- Using Elixir v1.11.3
- Using Erlang OTP v20.3
- With this code:
  ```elixir
    {:ok, document} = Floki.parse_document("
  <html>
    <style>
      .this-works {
        just: fine;
      }

      // Floki will think the p tag bellow is an actual element
      // Match all <p> tags
      p {
        color: red;
      }
    </style>
  </html>
  ")

  Floki.find(document, "style")
  |> IO.inspect(label: "Floki result")
  ```
Output is:
```
Floki result: [
  {"style", [],
   [
     "\n      .this-works {\n        just: fine;\n      }\n\n      // Floki will think the p tag bellow is an actual element\n      // Match all ",
     {"p", [], [" tags\n      p {\n        color: red;\n      }\n    "]}
   ]}
]
```

## Expected behavior

Floki should ignore html tags inside style tags. I did not check the problem with script tags so it it possible they have the same issue.

## Possible solution
https://github.com/philss/floki/blob/fd88a2836f4dd73d42c0d899d9aae5a132aaa7dc/lib/floki/html_tree.ex#L182
Add a pattern match to that function that matches `"style"` instead of `tag` and manually add a text node of the inner content instead of building a tree for the children of a style tag.

SweetMNM:
## Fixes #339

This is the fix for the issue.

At first I mentioned fixing the issue from the elixir side of things but I quickly realized it would be extremely ugly and probably inefficient.

So I hoped into the erlang code for mochiweb (Seeing erlang syntax for the first time was a bit overwhelming). After digging a bit I came across two function  `tokenize_script` and `tokenize_textarea`.

These functions provide the functionality we want, but for script and textarea tags. I wrote tests to make sure tags that are supposed to be plaintext get parsed correctly. And made them run on both style and script tags.  I was happy to see the tests only failed for the style tag but scripts are working just fine. So I created a function called `tokenize_style` which does what `tokenize_script` does but for style tags, and then implemented it.

✅ Tests passed

Great! The thing is there are more "plaintext" elements (eg title). I couldn't find a complete list of all of the elements.
The reason I did not implement a fix for the title tag as well is because the way mochiweb did it was to create a function for each tag (`tokenize_script`, `tokenize_textarea`). I am not sure why it did that instead of creating a function that works for all tags, perhaps for performance reasons. The thing I'm wondering is should I continue the tradition and create a function for each tag or try to create a function that works for all (which I am really not exited about since I can barely understand erlang)

Another thing is I made the new tests only run for Mochiweb since I could not get html5ever or fast_html to install/work properly, and since this PR is aimed at fixing Mochiweb this should not be a problem. Additionally no elixir code was changed, only Mochiweb code meaning the PR couldn't have broken html5ever/fasthtml functionality somehow.

### DIFF
--- a/test/floki_test.exs
+++ b/test/floki_test.exs
@@ -5,6 +5,11 @@ defmodule FlokiTest do
 
   alias Floki.HTMLParser.{Html5ever, Mochiweb, FastHtml}
 
+  @plain_text_tags [
+    "script",
+    "style"
+  ]
+
   @html """
   <html>
   <head>
@@ -140,6 +145,35 @@ defmodule FlokiTest do
                      {"o:p", [], []}
                    ]
       end
+    end
+
+    test "parse a HTML with tags that are plain text" do
+      validate_html = fn (tag) ->
+        {:ok, parsed} = tag
+        |> html_with_tag_that_should_not_have_children()
+        |> Floki.parse_document()
+
+        current_parser = Application.get_env(:floki, :html_parser)
+
+        case current_parser do
+          Mochiweb -> assert parsed ==
+              [
+                {"html", [],
+                [
+                  {"head", [], []},
+                  {"body", [],
+                    [
+                      {tag, [],
+                      ["this is not a <tag>\nthis is also </not> a tag\n and this is also not <a></a> tag"]}
+                    ]}
+                ]}
+              ]
+
+          _ -> {}
+        end
+      end
+
+      Enum.each(@plain_text_tags, validate_html)
     end
   end
 
@@ -1474,5 +1508,9 @@ defmodule FlokiTest do
         IO.inspect(x)
         raise "unexpected return from parser"
     end
+  end
+
+  defp html_with_tag_that_should_not_have_children(tag) do
+    html_body("<#{tag}>this is not a <tag>\nthis is also </not> a tag\n and this is also not <a></a> tag</#{tag}>")
   end
 end
--- a/test/floki_test.exs
+++ b/test/floki_test.exs
@@ -148,28 +148,33 @@ defmodule FlokiTest do
     end
 
     test "parse a HTML with tags that are plain text" do
-      validate_html = fn (tag) ->
-        {:ok, parsed} = tag
-        |> html_with_tag_that_should_not_have_children()
-        |> Floki.parse_document()
+      validate_html = fn tag ->
+        {:ok, parsed} =
+          tag
+          |> html_with_tag_that_should_not_have_children()
+          |> Floki.parse_document()
 
         current_parser = Application.get_env(:floki, :html_parser)
 
         case current_parser do
-          Mochiweb -> assert parsed ==
-              [
-                {"html", [],
-                [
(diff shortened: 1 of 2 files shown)


philss (approved):
@SweetMNM this looks great! Thank you! :purple_heart: 

> I am not sure why it did that instead of creating a function that works for all tags, perhaps for performance reasons. The thing I'm wondering is should I continue the tradition and create a function for each tag or try to create a function that works for all

I think you are correct. The reason is because each tag requires a specific binary match, as you implemented in `tokenize_style`. I can't see a way to implement a function that works for all.
My suggestion is to list the plain text tags and open a PR for each one of them.

> Another thing is I made the new tests only run for Mochiweb since I could not get html5ever or fast_html to install/work properly, and since this PR is aimed at fixing Mochiweb this should not be a problem. Additionally no elixir code was changed, only Mochiweb code meaning the PR couldn't have broken html5ever/fasthtml functionality somehow.

Cool! The CI is green, so it didn't broke indeed :)

:shipit:

SweetMNM:
@philss Sounds awesome! 😄
I will make this PR ready for review so you can merge it. 
After some digging I found [this](https://wicg.github.io/controls-list/html-output/multipage/syntax.html#html-fragment-parsing-algorithm) list. I am a bit confused by this list since it says:

>Set the state of the HTML parser's tokenization stage as follows, switching on the context element

but then it only says what context for some of the elements and some are just blank. Weird....

for now I create a different PR for the title tag or add it directly to this PR. Whatever you prefer.

test/floki_test.exs
```diff
@@ -141,6 +146,40 @@ defmodule FlokiTest do
...
+      validate_html = fn tag ->
+        {:ok, parsed} =
+          tag
+          |> html_with_tag_that_should_not_have_children()
+          |> Floki.parse_document()
+
+        current_parser = Application.get_env(:floki, :html_parser)
+
+        case current_parser do
+          Mochiweb ->
```
philss, line 160: For now this is good, but I think we could do something like a `tag` for tests that are specific to a configured parser.

philss (approved):
:shipit:

philss:
@SweetMNM thank you! :purple_heart:

test/floki_test.exs
SweetMNM: Agreed! At first I intended to assert different result for different parser since html5ever was parsing new lines differently. But then html5ever stopped working properly and I got tired of debugging it so we arrived at this. Will add a `tag` with the title PR

PR #343, opened 2021-04-03
Mochiweb - parse `title` tag as plaintext, Add `parser` ExUnit tags to skip tests based on current parser

SweetMNM:
Hello again 😀
Title is pretty self explanatory...

In addition to making Mochiweb parse title tags as plaintext, 2 new ExUnit tags were added:
`only_parser` and `except_parser`.
Quoted from the commit description:
> Due to some ExUnit tag limitations we can not use `@tag parser [Parser1, Parser2]`. (ExUnit does not allow tags to be list if they are excluded/included)
That's why the `except_parser` tag was created to run a test for all parsers except a specific one. This works fine when we have only 3 parsers...

> if in the future Floki has more than 3 parsers and a test needs to include only 2, I think the best approach would be to create a test macro that will be used like:

```elixir
test_for_parsers [Parser1, Parser2] do
  ...
end
```

### DIFF
--- a/test/floki_test.exs
+++ b/test/floki_test.exs
@@ -7,7 +7,9 @@ defmodule FlokiTest do
 
   @plain_text_tags [
     "script",
-    "style"
+    "style",
+    "title",
+    "textarea"
   ]
 
   @html """
--- a/test/floki_test.exs
+++ b/test/floki_test.exs
@@ -147,6 +147,7 @@ defmodule FlokiTest do
       end
     end
 
+    @tag only_parser: Mochiweb
     test "parse a HTML with tags that are plain text" do
       validate_html = fn tag ->
         {:ok, parsed} =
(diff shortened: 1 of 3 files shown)


test/floki_test.exs
```diff
@@ -147,35 +149,28 @@ defmodule FlokiTest do
       end
     end
 
+    @tag only_parser: Mochiweb
```
philss, line 152: First I got confused because it should be the entire name, but then I saw the alias. :smile:

philss (approved):
Looks great! Thank you! :purple_heart:

test/floki_test.exs
SweetMNM: Believe me I got confused as well but I just followed what you used for the `case` and it worked. Of course I later saw it was aliased 🤗

Issue #352, opened 2021-06-17
Support HTML ids which contain periods


nathanl:
## Feature goal

HTML ids which contain periods are legal, but Floki does not appear to support this. Periods would need to be escaped somehow in a selector so that they're not interpreted as classes.

For instance, to find `<form id="my.example.com-port-filters" class="foo">`, we might use `Floki.find(~s{form#my\.example\.com-port-filters.foo})`.

nathanl:
Turns out there's a workaround.

This came up in the context of using `LiveViewTest.element/3`. @chrismccord pointed out that the attribute syntax could work, which it does. And it also works here: `Floki.find(html, ~s{form[id="my.example.com-port-filters"]})`.

philss:
@nathanl I will investigate how we can escape the period. If you want to give it a try, I think we need to add a new rule for scaping in the lexer: https://github.com/philss/floki/blob/master/src/floki_selector_lexer.xrl
and then we need to parse it properly here: https://github.com/philss/floki/blob/master/lib/floki/selector/parser.ex

PR #353, opened 2021-06-17
Document selector for ids with periods

nathanl:
Document workaround for https://github.com/philss/floki/issues/352


philss:
@nathanl cool! Thank you! :purple_heart:

PR #359, opened 2021-10-05
Support HTML ids which contain periods

### LINKED ISSUE #352: Support HTML ids which contain periods

Hugo-Hache:
Add support for ids which contain periods.

Period in the selector is escaped in the same way as JavaScript's document.querySelector/1 (i.e. with a double backslash)

For instance, to find `<form id="my.example.com-port-filters" class="foo">`, we use `Floki.find(~s{form#my\\.example\\.com-port-filters.foo})`.

Fixes #352


Hugo-Hache:
👋 @philss, as I was looking to brush up my Elixir skills through hacktoberfest, I stumbled upon this issue. I hope you will find this contribution useful.

As I don't write a lot of Elixir, my code might not be very idiomatic (especially for the id replacement in `do_parse` arguments). Please let me know if you think it could be refactored more elegantly.

Hugo-Hache:
👋 @philss, as I was looking to brush up my Elixir skills through hacktoberfest, I stumbled upon this issue. I hope you will find this contribution useful.

As I don't write a lot of Elixir, my code might not be very idiomatic (especially for the id replacement in `do_parse` arguments). Please let me know if you think it could be refactored more elegantly.

Issue #399, opened 2022-03-28
html_declaration can be a 2 tuple


derek-zhou:
## Description

According to the [doc](https://hexdocs.pm/floki/Floki.html#t:html_declaration/0), html_declaration is a 3 tuple like:
```
@type html_declaration() :: {:pi, String.t(), [html_attribute()]}
```

However, I've encounter a `{:pi, String.t()}` 2 tuple. Reading the code a bit, this is actually intentional:

https://github.com/philss/floki/blob/d3fc8dada223eebfb1631164182e465d59836ca7/src/floki_mochi_html.erl#L351

`<?php` is broken html but unfortunately exists in the wild; most people would be happy if they do not exist. May I suggest to change this 2 tuple into a 3 tuple with a empty html_attribute list, so it is consistent with the documentation?

PR #400, opened 2022-03-28
remove case of 2 tuple of :pi

derek-zhou:
Like this? I didn't check the html5ever and fast_html parsers. 

https://github.com/philss/floki/issues/399


philss:
@derek-zhou I think so. Strangely we didn't treat that case in the selectors.

philss:
@derek-zhou thank you! :purple_heart:

PR #404, opened 2022-05-14
Minor stuff I changed while reading

hissssst:
* sufix -> suffix
* Fixed dialyzer types
* Minor optimizations for `Enum.find`
* Commented out unused functions

By the way, to find unused functions, you can use [this tool](https://hexdocs.pm/mix_unused/Mix.Tasks.Compile.Unused.html)


lib/floki/selector/combinator.ex
```diff
@@ -16,6 +16,12 @@ defmodule Floki.Selector.Combinator do
...
   defstruct match_type: nil, selector: nil
 
+  @type match_type :: nil
+  | :descendant
+  | :child
+  | :adjacent_sibling
+  | :general_sibling
+
   @type t :: %__MODULE__{
           match_type: :atom,
```
philss, line 26:
```suggestion
          match_type: match_type(),
```

lib/floki/selector/combinator.ex
```diff
@@ -16,6 +16,12 @@ defmodule Floki.Selector.Combinator do
 
   defstruct match_type: nil, selector: nil
 
+  @type match_type :: nil
```
philss, line 19: I guess it can't be nil. Can you confirm that?

lib/floki/html/tokenizer.ex
```diff
@@ -746,20 +746,21 @@ defmodule Floki.HTML.Tokenizer do
...
-      }
-    )
-  end
-
-  def script_data_escape_start(html, s) do
-    script_data(html, s)
-  end
+  ## Unused
+  # @spec script_data_escape_start(binary(), State.t()) :: State.t()
+  # def script_data_escape_start(<<?-, html::binary>>, s) do
```
philss, line 751: Thanks. There are some functions in this file that aren´t  in use. I left just for completion of the tokenizer spec... Maybe we should comment all of them. I can track them later.

philss (commented):
Thank you! :purple_heart: 

I added some suggestions. Also, please run `mix format`.

PR #410, opened 2022-07-11
Show retrieval of data attributes

nathanl:
Initially I wondered whether the attribute should be passed as `[data-name]` to match the selector.


philss:
@nathanl thank you! :purple_heart:

Issue #405, opened 2022-05-24
Suggestion self_closing_tags


inoas:
## Feature goal

Would you be interested in a PR that made these configurable. They would stay like this unless someone overwrote some application layer config.

https://github.com/philss/floki/blob/61cd7126c3486c4f857fc7f7a82405dcabf25df0/lib/floki/raw_html.ex#L4-L21

## Dependencies

None

inoas:
I am using Floki to render UBB-style xml and there are tags like `<pagelink/>` or `<media/>` or

philss:
@inoas this is good idea. Maybe `:custom_self_closing_tags`?

> If this interestes you: Would you also prefer for this to be changable at runtime like the html parser is?

This would be nice as well!

inoas:
OK, thanks for the go ahead, I will look to find some time to implement.

inoas:
> custom_self_closing_tags

The idea about `:append_self_closing_tags` and `:set_self_closing_tags` would be to either append to the ones currently set by default / by the developer... or to set them. `set_self_closing_tags` could also be `put_self_closing_tags` or `replace_self_closing_tags`.

This is up to you what you want here, just replacing or appending and replacing?
Just let me know :)

philss:
I personally don't see much use for replacing them entirely. Seems like only appending is enough.

PR #417, opened 2022-08-24
user configurable self closing tags

inoas:
Fixes [issue 405](https://github.com/philss/floki/issues/405).

Notice: Instead of just appending to the tag list, I have opted to expose the internal default tag list and replace it.

The reason for this is that in some specific xml dialects people would want to write they could replace `` with `...` and floki would not be able to support that.

The trade-off here in terms of DX is that if you want to keep the standard html self closing tags you cannot just do:

```elixir
config :floki, :self_closing_tags, ["download"]
```

but you need to do

```elixir
config :floki, :self_closing_tags, ["tags", "you", "want"]
```

I am willing to document this and I hope that this solution passes your checks.

Please let me know if the way I have added unit tests is the the right approach for you.

p.s.: I have tested the feature in my small real world application where I am using a custom XMl dialect to allow a subset of html and extend it by some specific tags of which some are self-closing:

```elixir
# In my userland application `config/config.exs`
config :floki, :self_closing_tags, [
  "article",
  "br",
  "download",
  "hr",
  "media",
  "page",
  "shy",
  "space",
  "wbr"
]
```

### DIFF
--- a/lib/floki/raw_html.ex
+++ b/lib/floki/raw_html.ex
@@ -1,7 +1,7 @@
 defmodule Floki.RawHTML do
   @moduledoc false
 
-  @self_closing_tags [
+  @default_self_closing_tags [
     "area",
     "base",
     "br",
@@ -13,12 +13,23 @@ defmodule Floki.RawHTML do
     "input",
     "keygen",
     "link",
+    "menuitem",
     "meta",
     "param",
     "source",
     "track",
     "wbr"
   ]
+
+  def default_self_closing_tags(), do: @default_self_closing_tags
+
+  def self_closing_tags do
+    custom_self_closing_tags = Application.get_env(:floki, :self_closing_tags)
+
+    if is_list(custom_self_closing_tags),
+      do: custom_self_closing_tags,
+      else: @default_self_closing_tags
+  end
 
   @encoder &HtmlEntities.encode/1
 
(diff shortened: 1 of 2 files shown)


inoas:
Further notes:

1. Once this is GTM, I can squash this to remove the noise.
2. I have opted to let `Floki.RawHTML.default_self_closing_tags` and `Floki.RawHTML.self_closing_tags` be public so that user land applications can list and/or document self closing tags programatically.
3. I tried this:
   ```elixir
   config :floki, :self_closing_tags, Floki.RawHTML.default_self_closing_tags() ++ ["br"]
   ```
   ... in `config/config.exs` but at that stage the module is not compiled/available yet. I am open to suggestions to make it easier to extend the list of self closing tags instead of overwriting them.

inoas:
I have added `menuitem` because we also support `command` and `keygen`, all 3 are obsolete but that doesn't mean that they don't exist in the wild, so by default I'd include all 3.

philss:
@inoas I was thinking maybe we should use [`Application.compile_env/3`](https://hexdocs.pm/elixir/1.13.4/Application.html#compile_env/3) because it would be faster and the code could stay almost the same as before (using module tag). The only drawback would be testing, which is not possible since it's compiled I think. But I guess this is fine.  WDYT?

philss:
@inoas and BTW, thanks for the PR! :purple_heart:

inoas:
> @inoas I was thinking maybe we should use [`Application.compile_env/3`](https://hexdocs.pm/elixir/1.13.4/Application.html#compile_env/3) because it would be faster and the code could stay almost the same as before (using module tag). The only drawback would be testing, which is not possible since it's compiled I think. But I guess this is fine. WDYT?

Do you mean like here? https://github.com/philss/floki/pull/417/files#diff-477200ed1da650829519cb2d3f87fdd9a8b3d5f8d786993494dbbeceb795445eR27

I mean sure I can switch this around and remove the tests.
But I don't think the changes I did will really impact performance, do you have some tools to measure?

What are you concerned about: the `get_env` call for every tag or the move from guards to `case`s because I don't think that the latter will have a huge impact (but the former may).

If there are ways to do this the right way (tested + fast) via macros and you have an idea, just point me into the direction and I will see what I can do about it.

philss:
@inoas let's keep it simple and go with your current version. We can iterate later and see if there is any gain to use the `compile_env`.

inoas:
please trigger ci

lib/floki/raw_html.ex
```diff
@@ -99,13 +110,19 @@ defmodule Floki.RawHTML do
...
+      {true, []} -> "/>"
+      _ -> ">"
+    end
+  end
 
-  defp close_end_tag(type, _, padding),
-    do: [leftpad(padding), "</", type, ">", line_ending(padding)]
+  defp close_end_tag(type, children, padding) do
+    case {type in self_closing_tags(), children} do
+      {true, []} -> ""
```
philss, line 122: It's slightly cheaper to have an empty list than an empty string. Since we are using IO data here, this is possible:

```suggestion
      {true, []} -> []
```

philss (approved):
Thank you! :purple_heart: 

I sent a suggestion, but I'm going to commit it already.

PR #431, opened 2022-10-30
Fix traverse_and_update/3 spec

### LINKED ISSUE #426: traverse_and_update/3 return typespec needs html_tree()
## Description

The first element in the tuple return type was changed to`html_node()` in this revision - https://github.com/philss/floki/pull/289/files#diff-0eacaeb3de3bdf5b80408719a3ce69f47fd9eda5e9a02de33bd65f7b1235e0e2R395

But dialyzer caught for me that this was always returning an `html_tree()` spec'd as a list with a root node -- 
`@type html_tree() :: [html_node()]` 

I don't pretend to know if maybe other use cases might return just a node - but if that's the case could we change spec for return type to allow `html_tree() | html_node()` node?  Thanks in advance for taking the time to read my report & for the awesome lib!

## To Reproduce

- Using Floki v0.33.1
- Using Elixir v1.14.0-otp-25
- Using Erlang OTP v25.0.3
- With this code:
Here's a simple reproduction:

```elixir
iex(27)> simple = "<div></div>"
"<div></div>"
iex(28)> Floki.parse_document!(simple) |> Floki.traverse_and_update(0, fn node, acc -> {node, acc} end)
{[{"div", [], []}], 0}
```

## Expected behavior

Based on typespec, I expected:
```diff
- {[{"div", [], []}], 0}
+ {{"div", [], []}, 0}
```

WLSF:
Fix #426

### DIFF
--- a/lib/floki.ex
+++ b/lib/floki.ex
@@ -403,9 +403,9 @@ defmodule Floki do
   """
 
   @spec traverse_and_update(
-          html_tree(),
-          (html_tag() | html_comment() | html_doctype() | html_declaration() -> html_node() | nil)
-        ) :: html_tree()
+          html_node() | html_tree(),
+          (html_node() -> html_node() | nil)
+        ) :: html_node() | html_tree()
 
   defdelegate traverse_and_update(html_tree, fun), to: Floki.Traversal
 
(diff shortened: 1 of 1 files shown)


WLSF:
@philss I guess it should be good now!

philss:
@WLSF thank you! :purple_heart:

PR #435, opened 2022-11-04
Add unused dependencies check step to CI

antedeguemon:
Hey there! 

This is a simple quality-of-life check to ensure no unused dependencies are locked by `mix.lock`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,6 +41,9 @@ jobs:
       - name: Install dependencies
         run: mix deps.get
 
+      - name: Check for unused dependencies
+        run: mix deps.unlock --check-unused
+
       - name: Install Rust toolchain
         uses: actions-rs/toolchain@v1
         with:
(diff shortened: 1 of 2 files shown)


philss:
@antedeguemon thank you for the PR! :purple_heart: 
I think this won't work because we have different dependencies for each parser.

Perhaps we could declare all of them normally in the `mix.exs` file, since they don't take too much time to build now. Can you give it a try?

philss (approved):
Cool! Much better now! :shipit:

antedeguemon:
@philss yay! I've removed the PARSER environment variable and its related logic.

CI time stayed pretty close:
- [Before this PR: 1m 26s](https://github.com/philss/floki/actions/runs/3387969590)
- [After this PR: 1m 39s](https://github.com/philss/floki/actions/runs/3397106451)

philss:
@antedeguemon Nice! I noticed that we are installing the Rust toolchain, but I think this is not necessary anymore, since `html5ever` is using a precompiled binary. If you want to remove that from the workflow, feel free to send a PR :) (otherwise I can take this later).

PR #437, opened 2022-11-17
Fix dialyzer issues with Floki.text/1

davydog187:
If you pass a node directly, you'll get an error like

```
The function call will not succeed.

Floki.text(_block :: {<<_::8>>, binary() | [{binary(), binary()}], [any()]})

breaks the contract
(html_tree() | binary()) :: binary()
```

### DIFF
--- a/lib/floki.ex
+++ b/lib/floki.ex
@@ -502,7 +502,7 @@ defmodule Floki do
 
   """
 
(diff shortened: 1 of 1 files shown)


philss:
@davydog187 thank you! :purple_heart:

davydog187:
Thanks for the fast merge @philss !

PR #446, opened 2023-02-13
Pass arguments to parser modules

Kuret:
Hi there!

For testing partials with `fast_html`, we needed to pass arguments to its `parse` method.
Since Floki did not support passing arguments to their HTML Parsers an option to pass arguments to the parser module was added.

Note that this only has effect on `fast_html`, since both `mochiweb` and `html5ever` do not support additional arguments.

**Why we needed this option:**
This was needed because (for example) we had a partial for a table row. When testing this partial on it's own, the output would ignore the `tr` and `td` element, since it was 'invalid html' as the parent `table` element was missing.

This can be solved by passing a `context: "table"` to `fast_html`'s parse function, only Floki did not support passing along arguments.

For example:
```
Floki.parse_fragment("<tr><td>Column 1</td><td>Column 2</td></tr>")
> {:ok. ["Column 1Column2"]}

Floki.parse_fragment("<tr><td>Column 1</td><td>Column 2</td></tr>", parser_args: [context: "table"])
> {:ok. [{ "tbody", [], [{"tr", [], [{"td", [], ["1"]}, {"td", [], ["2"]}]}]}]}
```

### DIFF
--- a/lib/floki/html_parser.ex
+++ b/lib/floki/html_parser.ex
@@ -21,15 +21,19 @@ defmodule Floki.HTMLParser do
 
   @default_parser Floki.HTMLParser.Mochiweb
 
-  @callback parse_document(binary()) :: {:ok, Floki.html_tree()} | {:error, String.t()}
-  @callback parse_fragment(binary()) :: {:ok, Floki.html_tree()} | {:error, String.t()}
+  @callback parse_document(binary(), list()) :: {:ok, Floki.html_tree()} | {:error, String.t()}
+  @callback parse_fragment(binary(), list()) :: {:ok, Floki.html_tree()} | {:error, String.t()}
 
   def parse_document(html, opts \\ []) do
-    parser(opts).parse_document(html)
+    parser_args = opts[:parser_args] || []
+
+    parser(opts).parse_document(html, parser_args)
   end
 
   def parse_fragment(html, opts \\ []) do
-    parser(opts).parse_fragment(html)
+    parser_args = opts[:parser_args] || []
+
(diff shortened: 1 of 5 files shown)


philss:
Hi @Kuret :wave: 

Thank you for the PR. I liked the idea, but I think we could have a different API.

It's common in Elixir that when we want to pass options to a dependency that is configurable, we use the `MFA` or the `MA` approach. `MFA` stands for "Module, function and arguments", and is generally presented like: `{ModuleName, :function, [args]}`.

My idea is to accept the parser configuration in the `MA` way: `{ModuleName, [parse_args]}`.
So you would instead define an application config **or** use your function like this:

```elixir
Floki.parse_fragment("<tr><td>Column 1</td><td>Column 2</td></tr>", parser: {CustomParser, [context: "table"]})
```

WDYT?

Kuret:
Hi,

Thanks for your reply.

Sounds good and is probably a more ‘Elixir’ way to implement it indeed.

Kuret:
When giving it another thought, it does sound like it’s easier to just pass the options when needed instead of passing the Module (which I already configured globally probably) with options every time a simple option is needed.

Also, some of these options are function specific and only work on ‘fragment’ for example, and now it looks like the options are passed on the Module level.

Seems a bit redundant just to do it in a more ‘Elixir’ way?

philss (approved):

philss:
@Kuret makes sense. I think we can support both ways, and maybe merge the args if they are passed both as MA and `:parser_args`.

Thank you! :purple_heart:

PR #447, opened 2023-02-21
Support splitting nodes while traversin

martosaur:
Hi! This PR adds support for returning a list of nodes in `Floki.traverse_and_update`.

The use case for this is if we want to replace an inline element with a block-level one, i.e.

```html
<p>Code: <code>IO.inspect</code></p>
```
->
```html
<p>Code:</p>
<pre>IO.inspect</pre>
```

### DIFF
--- a/lib/floki/traversal.ex
+++ b/lib/floki/traversal.ex
@@ -21,6 +21,7 @@ defmodule Floki.Traversal do
 
       {mapped_head, new_acc} ->
         {mapped_tail, new_acc2} = traverse_and_update(tail, new_acc, fun)
+
         mapped =
           if is_list(mapped_head) do
             mapped_head ++ mapped_tail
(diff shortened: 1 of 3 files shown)


lib/floki/traversal.ex
```diff
@@ -21,7 +21,7 @@ defmodule Floki.Traversal do
 
       {mapped_head, new_acc} ->
         {mapped_tail, new_acc2} = traverse_and_update(tail, new_acc, fun)
-        {[mapped_head | mapped_tail], new_acc2}
+        {List.flatten([mapped_head | mapped_tail]), new_acc2}
```
philss, line 24: Could we match for the sides and use a concat `++` instead of a flatten?

philss: *when needed

philss:
```suggestion
        mapped =
          if is_list(mapped_head) do
            mapped_head ++ mapped_tail
          else
            [mapped_head | mapped_tail]
          end

        {mapped, new_acc2}
```

philss:
@martosaur thank you! :purple_heart:

PR #458, opened 2023-05-19
Parse escaped colons in class names correctly

### LINKED ISSUE #411: Floki.find doesn't support escaped colons in class names
## Description

Floki doesn't support selectors with escaped colons in them, probably it's mixing them up with pseudo selectors

## To Reproduce

- Using Floki v0.33.1
- Using Elixir v1.13.3
- Using Erlang OTP v24
- With this code:

```elixir
 > Floki.find(document, "a.xs\\:red-500")
[] 
```

## Expected behavior
```elixir
 > Floki.find(document, "a.xs\\:red-500")
[{"a", [{"class", "xs\\:red-500"}], ["I'm a link with a funny selector"]}]
```

The current workaround is to use `a[class="xs:red-500"]`

viniciusmuller:
This pull requests adds an extra rule in the parser in order to prevent escaped class names (`a.xs\\:red-500`) from being parsed as  pseudo selectors.

Closes #411

### DIFF
--- a/test/floki/selector/parser_test.exs
+++ b/test/floki/selector/parser_test.exs
@@ -22,6 +22,14 @@ defmodule Floki.Selector.ParserTest do
     end
   end
 
+  test "escaped colons in class names" do 
+    tokens = tokenize("a.xs\\:red-500")
+
+    assert Parser.parse(tokens) == [
+      %Selector{type: "a", classes: ["xs\\:red-500"], pseudo_classes: []}
+    ]
+  end
+
   test "multiple selectors" do
     tokens = tokenize("ol, ul")
 
--- a/test/floki/selector/parser_test.exs
+++ b/test/floki/selector/parser_test.exs
@@ -22,12 +22,12 @@ defmodule Floki.Selector.ParserTest do
     end
   end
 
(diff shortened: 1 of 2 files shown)


philss:
@viniciusmuller thank you!! :purple_heart:

test/floki/selector/parser_test.exs
```diff
@@ -22,6 +22,14 @@ defmodule Floki.Selector.ParserTest do
     end
   end
 
+  test "escaped colons in class names" do
+    tokens = tokenize("a.xs\\:red-500")
+
+    assert Parser.parse(tokens) == [
+             %Selector{type: "a", classes: ["xs\\:red-500"], pseudo_classes: []}
```
viniciusmuller, line 29: @philss I was thinking here, do you think that this is the expected behavior? Or we actually want to return "xs:red-500", I did this based on the original issue, but maybe the author made a typo in the example code

philss: @viniciusmuller I did a quick fix in #464 

I believe the author was trying to find elements with Tailwind classes, which can have colons in it.

PR #459, opened 2023-05-19
Add `include_inputs?` option to Floki.text

### LINKED ISSUE #391: Include input values in the result of `Floki.text/2`
## Feature goal

Include the values of `<input ...>` elements in `Floki.text/2`.

This would be helpful for making assertions on pages that have edit forms:
```elixir
test "the form is prefilled with the current entity" do
  # ...
  text = Floki.text(html)

  assert text =~ "current value"
end
```

This could be implemented as an option for `Floki.text/2`:

```
Floki.text(html, inputs: true)
```
or
```
Floki.text(html, inputs: ["text"])
```

viniciusmuller:
This pull request adds the `include_inputs?` option to the `Floki.text` function.

Example usage:
```elixir
iex(1)> Floki.text([{"input", [{"type", "date"}, {"value", "2017-06-01"}], []}], include_inputs?: true)
"2017-06-01"
```

Currently it will add into the resulting text all the values from a list of allowed inputs (excluding things such as checkboxes and radios)

Feel free to review and point any changes you would like @philss

Closes #391

### DIFF
--- a/lib/floki/deep_text.ex
+++ b/lib/floki/deep_text.ex
@@ -6,25 +6,35 @@ defmodule Floki.DeepText do
 
   @type html_tree :: tuple | list
 
-  @spec get(html_tree, binary) :: binary
+  @spec get(html_tree, binary, boolean) :: binary
 
-  def get(html_tree, sep \\ "") do
-    get_text(html_tree, "", sep)
+  def get(html_tree, sep \\ "", include_inputs? \\ false)
+
+  def get(html_tree, sep, include_inputs?) do
+    get_text(html_tree, "", sep, include_inputs?)
   end
 
-  defp get_text(text, "", _sep) when is_binary(text), do: text
-  defp get_text(text, acc, sep) when is_binary(text), do: Enum.join([acc, text], sep)
+  defp get_text(text, "", _sep, _) when is_binary(text), do: text
+  defp get_text(text, acc, sep, _) when is_binary(text), do: Enum.join([acc, text], sep)
 
-  defp get_text(nodes, acc, sep) when is_list(nodes) do
+  defp get_text(nodes, acc, sep, include_inputs?) when is_list(nodes) do
     Enum.reduce(nodes, acc, fn child, istr ->
-      get_text(child, istr, sep)
+      get_text(child, istr, sep, include_inputs?)
     end)
   end
 
-  defp get_text({:comment, _}, acc, _), do: acc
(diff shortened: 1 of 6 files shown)


lib/floki/deep_text.ex
```diff
@@ -6,25 +6,35 @@ defmodule Floki.DeepText do
 
   @type html_tree :: tuple | list
 
-  @spec get(html_tree, binary) :: binary
+  @spec get(html_tree, binary, boolean) :: binary
 
-  def get(html_tree, sep \\ "") do
-    get_text(html_tree, "", sep)
+  def get(html_tree, sep \\ "", include_inputs? \\ false)
```
philss, line 11: Here I think it's OK to keep the `?` in the variable name.

lib/floki/text_extractor.ex
```diff
@@ -0,0 +1,42 @@
+defmodule Floki.TextExtractor do
+  @moduledoc false
+
+  @allowed_input_types [
+    nil,
```
philss, line 5: Why `nil` is allowed?

lib/floki/text_extractor.ex
```diff
@@ -0,0 +1,42 @@
...
+
+    case type do
+      {"type", t} ->
+        if t in @allowed_input_types do
+          extract_value(attrs)
+        else
+          ""
+        end
+
+      nil ->
```
philss, line 31: If we couldn't find the input type, shouldn't we just return `nil`? Or `nil` type implies that this is a `text` input? :thinking:

philss (commented):
Hey @viniciusmuller :wave: 

Thank you for the PR! :purple_heart:  I added some comments to it.
But overall it looks good!

lib/floki/text_extractor.ex
viniciusmuller: That's for inputs that lack the `type` attribute, since  they default to `text`.

But looking now I think it will be better to make the input type search default "text" in the `extract_input_values` function.

What do you think?

philss: Yeah, makes more sense to me :+1:

lib/floki/text_extractor.ex
viniciusmuller: I've simplified this part to default an input's type to `text` if it doesn't have the attribute, I believe the code is clearer now

philss (approved):
:shipit:

Issue #463, opened 2023-05-31
Allow option to parse attributes as maps


chrismccord:
With the OTP 26 map optimizations, it has become clear that I was matching on incidentally generated html attributes in specific order. This is a problem to test against, because Floki parses the attributes in the order they appear into a list, so if Phoenix or LiveView starts spitting out attributes in different order, tests fail even tho we are relying on HTML parsing to avoid matching directly on strings. Parsing attributes into a map will allow equality matching on the markup AST in tests and prevents ordering issues. What do you think? Thanks!

philss:
@chrismccord So the idea is to add an option to [`Floki.parse_document/2`](https://hexdocs.pm/floki/Floki.html#parse_document/2) and [`Floki.parse_fragment!/2`](https://hexdocs.pm/floki/Floki.html#parse_document/2), right?
I think this is tough because it would require the change of some functions, but I think it's possible, and it's a good addition! 

PS: I think we will not have this problem once we implement https://github.com/philss/floki/issues/457, but that would be a bigger change. cc @wojtekmach

bennelsonweiss:
Correct me if I'm misunderstanding your proposal, but parsing attributes into a map would lose data (the attribute ordering).

The loss of this data might manifest in ways such as the output of rendering parsed HTML being different than the originally parsed HTML. Is the benefit of equality matching AST really worth losing potentially useful data in the process?

In regards to changes to the order of rendered attributes in Phoenix or LiveView, shouldn't attributes be rendered in a consistent fashion to avoid continuously altering the produced HTML (helping cacheability, etc)?

chrismccord:
Yes exactly. And yes, it will be problematic for a number of areas, such as `Floki.raw_html` and `Floki.text` needing to become map aware (which shouldn't be a massive lift). I'm sure there are other implications as well, but I think it's essential going forward to test HTML output without relying on attribute ordering.

chrismccord:
> Correct me if I'm misunderstanding your proposal, but parsing attributes into a map would lose data (the attribute ordering).
> 
> The loss of this data might manifest in ways such as the output of rendering parsed HTML being different than the originally parsed HTML. Is the benefit of equality matching AST really worth losing potentially useful data in the process?
> 
> In regards to changes to the order of rendered attributes in Phoenix or LiveView, shouldn't attributes be rendered in a consistent fashion to avoid continuously altering the produced HTML (helping cacheability, etc)?

[...]

PR #467, opened 2023-06-09
Add support for parsing attributes as maps

philss:
This is a feature request from https://github.com/philss/floki/issues/463 and is the base for implementing the same feature in more backends.

Right now only the built-in `mochiweb` parser implements the feature.

### DIFF
--- a/lib/floki.ex
+++ b/lib/floki.ex
@@ -63,12 +63,14 @@ defmodule Floki do
   inside a list.
   """
 
-  @type html_declaration :: {:pi, String.t(), [html_attribute()]}
+  @type html_attribute :: {String.t(), String.t()}
+  @type html_attributes :: [html_attribute()] | html_attributes_map()
+  @type html_attributes_map :: %{String.t() => String.t()}
+  @type html_declaration :: {:pi, String.t(), html_attributes()}
   @type html_comment :: {:comment, String.t()}
   @type html_doctype :: {:doctype, String.t(), String.t(), String.t()}
-  @type html_attribute :: {String.t(), String.t()}
   @type html_text :: String.t()
-  @type html_tag :: {String.t(), [html_attribute()], [html_node()]}
+  @type html_tag :: {String.t(), html_attributes(), [html_node()]}
   @type html_node ::
           html_tag() | html_comment() | html_doctype() | html_declaration() | html_text()
   @type html_tree :: [html_node()]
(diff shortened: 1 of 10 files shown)


lib/floki.ex
```diff
@@ -102,13 +104,20 @@ defmodule Floki do
 
   ## Options
 
+    * `:attributes_as_maps` - Change the behaviour of the parser to return the attributes
+      as maps, instead of a list of `{"key", "value"}`. Remember that maps are no longer 
+      ordered since OTP 26. Default to `false`.
```
josevalim, line 109: Maps were never ordered (rather, they never guaranteed a particular ordering). :)

```suggestion
      as maps, instead of a list of `{"key", "value"}`. Default to `false`.
```

lib/floki.ex
```diff
@@ -152,13 +168,20 @@ defmodule Floki do
...
+
     * `:html_parser` - The module of the backend that is responsible for parsing
       the HTML string. By default it is set to the built-in parser, and the module
       name is equal to `Floki.HTMLParser.Mochiweb`, or from the value of the
       application env of the same name.
 
       See https://github.com/philss/floki#alternative-html-parsers for more details.
 
+    * `:parser_args` - A list of options to the parser. This can be used to pass options
+      that are specific for a given parser. Defaults to an empty list.
```
josevalim, line 183: I am curious, what would be the use case for this option?

philss: This was introduced to change the behaviour of a specific parser. Like for the `FastHTML`, that we can pass a `:context` option to it - we do that here: https://github.com/philss/floki/blob/2d5c282099764f12a21ea37d644b9ea09971d0eb/test/floki_test.exs#L242.

It was undocumented before, so I added a mention about it.

josevalim: Perfect!

src/floki_mochi_html.erl
```diff
@@ -215,38 +226,44 @@ tree_data([{data, Data, Whitespace} | Rest], AllWhitespace, Acc) ->
...
-    {norm(Tag), [{norm(K), iolist_to_binary(V)} || {K, V} <- Attrs], []};
-norm(Tag) when is_binary(Tag) ->
+tree([{doctype, _} | Rest], Stack, Opts) ->
+    tree(Rest, Stack, Opts).
+
+norm({Tag, Attrs}, Opts) ->
+    Attrs = [{norm(K, Opts), iolist_to_binary(V)} || {K, V} <- Attrs],
+    case lists:keyfind(attributes_as_maps, 1, Opts) of
+        {attributes_as_maps, true} ->
+            {norm(Tag, Opts), maps:from_list(Attrs), []};
```
bennelsonweiss, line 260: Documentation on `maps:from_list` says

> If the same key appears more than once, the latter (right-most) value is used and the previous values are ignored.

This means for duplicate attributes the _last_ attribute is used in the map.

The HTML spec suggests that it's a parse error when you encounter multiple identical attributes ([duplicate-attribute parse error](https://html.spec.whatwg.org/#parse-error-duplicate-attribute)), and that the parser ignores the duplicates (keeping the first).

philss: Great catch! Fixed in 708ad4406012097d1514dee328f1f08ad4ccefe3
Thank you! :purple_heart:

PR #497, opened 2023-10-31
`find/2` input protection proposal.

Hajto:
Today I had an issue where I accidentally passed `{:ok, tree}` into the `find/2`. I think it might not be a bad idea to protect the input with a guard. People who can't read the docs like me might save some frustration.

This is purely proposal. If I get a green light for this I will make develop this PR properly.

### DIFF
--- a/lib/floki.ex
+++ b/lib/floki.ex
@@ -78,9 +78,8 @@ defmodule Floki do
   @type css_selector :: String.t() | %Floki.Selector{} | [%Floki.Selector{}]
 
   defguard is_html_node(value)
-           when is_binary(value) or
-           is_list(value) or tuple_size(value) == 3 or
-                  (tuple_size(value) == 2 and elem(value, 0) in [:pi, :comment, :doctype])
+           when is_binary(value) or tuple_size(value) == 3 or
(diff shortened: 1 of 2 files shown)


lib/floki.ex
```diff
@@ -1,4 +1,5 @@
 defmodule Floki do
+  require Floki
```
philss, line 2: This is not needed :)
```suggestion
```

philss (commented):
I think it's a good idea. Please go ahead and try to fix the tests :)

lib/floki.ex
Hajto: Sorry, some sort of brain fart.

Hajto:
image
Should this be deleted?

Hajto:
Added: 
 - Tests
Fixed:
 -  wrong doctype specification (it looks like it will never affect the `find/2` guard but fixed for completenes)
 -  moved list match out of guard definition since list is not valid node

lib/floki.ex
Hajto: VS Code added that line accidentally. I removed it. Good catch.

philss:
@Hajto thank you! :purple_heart:

PR #498, opened 2023-11-05
Update raw_html.ex to handle :attributes_as_maps option

SupaMic:
When parsing using the new option :attributes_as_maps, I got errors when trying to convert back with the raw_html function.  This changes the Map into a list where the error occurs and continues regular processing.


philss (commented):
Good call! Would you mind to add a test case to it?

SupaMic:
I added a test which I think covers a basic use case but let me know if you think we should test more

philss:
@SupaMic thank you! :purple_heart:

PR #500, opened 2023-11-20
Remove outdated gleam wrapper code

michallepicki:
The external functions syntax in Gleam has changed: https://gleam.run/book/tour/external-functions.html

There's also no easy way to use this code from a gleam project that just has the floki dependency added with the gleam build tool. I think it's safe to say that no one is using this. Likely a better way to wrap Floki in Gleam would be to make a separate Gleam package with floki as a dependency, and wrap the API there - maybe then the package could also provide a more Gleam-y API than "here's an opaque thing, decode it if you want"


philss:
@michallepicki Makes sense! Thank you! I'm going to assume that no one is using.

If anyone needs a wrapper, please go ahead and create a new package.

PR #504, opened 2023-12-15
Enable parse of IO data html

ypconstante:
`parse_document` and `parse_fragment` declare in their spec that they support `iodata()`, but this format is not actually working today. 
This PR adds support to IO data on all parsers, which is useful specially on mochiweb, since it has to add a `<floki>` tag around the html.


philss:
Thank you!

PR #507, opened 2023-12-19
Restrict `HTMLParser` interface to only accept binaries

philss:
This makes the interfaces simplier. We don't need to support iodata for now yet.


philss:
cc/ @wojtekmach

ypconstante:
`iodata` support would be useful on Phoenix LiveView tests, for example. [Today Phoenix  has the content stored as iodata and converts to binary just to call parse](https://github.com/phoenixframework/phoenix_live_view/blob/main/lib/phoenix_live_view/test/dom.ex#L277-L279), and because `mochi_html` adds the `<floki>` root element, for each parse we ended up with two identical binaries being created.

philss:
OK, I will keep as it is. Thanks for pointing out, @ypconstante

PR #508, opened 2023-12-21
Use recursion instead of Enum.flat_map on find_selectors

ypconstante:
Today `find_selectors`  and it's child functions use `Enum.flat_map` to interact through the nodes to find the matching ones.

This PR replaces the `Enum.flat_map` and `Enum.map`  usages with recursion, removing most of the intermediate lists this methods had, and also simplifying `traverse_with` implementation.

This also provides a really small performance improvement. When checking this benchmark with fprof, operations count went from 65867 to 57726, reducing the time to run with fprof from 115ms to 95ms.

```
Name            ips        average  deviation         median         99th %
pr           555.12        1.80 ms    ±13.53%        1.72 ms        2.75 ms
today        544.31        1.84 ms    ±12.83%        1.75 ms        2.79 ms

Comparison: 
pr           555.12
today        544.31 - 1.02x slower +0.0358 ms

Memory usage statistics:

Name     Memory usage
pr            2.05 MB
today         2.09 MB - 1.02x memory usage +0.0319 MB

```

```elixir 
read_file = fn name ->
  __ENV__.file
  |> Path.dirname()
  |> Path.join(name)
  |> File.read!()
end

html_input = read_file.("small.html")

Benchee.run(
  %{
    "bench" => fn -> Floki.Finder.find(html, ".mw-parser-output > p")  end
  },
  time: 5,
  memory_time: 2
)
```

### DIFF
--- a/lib/floki/finder.ex
+++ b/lib/floki/finder.ex
@@ -45,8 +45,11 @@ defmodule Floki.Finder do
     results =
       tree.node_ids
       |> Enum.reverse()
-      |> get_nodes(tree)
-      |> Enum.flat_map(fn html_node -> get_matches_for_selectors(tree, html_node, selectors) end)
+      |> Enum.reduce([], fn node_id, acc ->
+        html_node = get_node(node_id, tree)
+        get_matches_for_selectors(tree, html_node, selectors, acc)
+      end)
+      |> Enum.reverse()
       |> Enum.uniq()
 
     {tree, results}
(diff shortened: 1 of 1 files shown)


philss (approved):

philss:
I have to say it was hard to remember this part of the code :sweat_smile: 
Nicely done! Thank you! :purple_heart:

PR #509, opened 2023-12-21
Use list instead of regex for string split

ypconstante:
Today the class and attribute selector are using String.split with a regex, but [using a list of separators instead of regex is faster](https://github.com/devonestes/fast-elixir#splitting-large-strings-code).

```
Name                    ips        average  deviation         median         99th %
pr                   668.99        1.49 ms    ±15.56%        1.43 ms        2.39 ms
today                528.96        1.89 ms    ±13.51%        1.81 ms        2.84 ms

Comparison: 
pr                   668.99
today                528.96 - 1.26x slower +0.40 ms

Memory usage statistics:

Name             Memory usage
pr                    1.92 MB
today                 2.09 MB - 1.08x memory usage +0.161 MB
```
```
read_file = fn name ->
  __ENV__.file
  |> Path.dirname()
  |> Path.join(name)
  |> File.read!()
end

html_input = read_file.("small.html")

Benchee.run(
  %{
    "bench" => fn -> Floki.Finder.find(html, ".mw-parser-output > p")  end
  },
  time: 5,
  memory_time: 2
)
```

### DIFF
--- a/lib/floki/selector.ex
+++ b/lib/floki/selector.ex
@@ -144,7 +144,7 @@ defmodule Floki.Selector do
   defp do_classes_matches?(nil, _classes), do: false
 
   defp do_classes_matches?(class_attr_value, classes) do
-    classes -- String.split(class_attr_value, [" ", "\t"]) == []
+    classes -- String.split(class_attr_value, [" ", "\t", "\n"]) == []
   end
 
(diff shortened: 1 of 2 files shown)


ypconstante:
Replaced by https://github.com/philss/floki/pull/510

PR #510, opened 2023-12-22
Optimize selectors matching

ypconstante:
Apply some optimizations on selector matching to try to avoid expensive calls:
- Use list instead of regex for String.split - [benchmark](https://github.com/philss/floki/pull/509)
- Don't try to match class if attribute value bit_size is smaller than the first class in the selector
- Reorder selector classes to put longest classes first, to optimize the check above
- Use equals when there's a single class selector and the class attribute value has the same bit_size as the selector
- Compute minimal class attribute value bit_size, and avoid split if this condition isn't match 
- on `match?`, check if selector requires descendant, and skip other checks if the current node doesn't have descendants

```
##### With input child #####
Name                 ips        average  deviation         median         99th %
bench (pr)        103.12        9.70 ms    ±18.72%        9.61 ms       14.42 ms
bench              81.41       12.28 ms    ±18.12%       12.18 ms       18.23 ms

Comparison: 
bench (pr)        103.12
bench              81.41 - 1.27x slower +2.59 ms

Memory usage statistics:

Name          Memory usage
bench (pr)        10.83 MB
bench             11.63 MB - 1.07x memory usage +0.80 MB

**All measurements for memory usage were the same**

##### With input descendant #####
Name                 ips        average  deviation         median         99th %
bench (pr)        115.90        8.63 ms    ±18.97%        7.92 ms       13.61 ms
bench              91.88       10.88 ms    ±17.60%       10.12 ms       16.85 ms

Comparison: 
bench (pr)        115.90
bench              91.88 - 1.26x slower +2.26 ms

Memory usage statistics:

Name          Memory usage
bench (pr)        10.81 MB
bench             11.55 MB - 1.07x memory usage +0.73 MB

**All measurements for memory usage were the same**

##### With input multiple classes #####
Name                 ips        average  deviation         median         99th %
bench (pr)        138.85        7.20 ms    ±15.04%        6.70 ms       10.50 ms
bench              96.60       10.35 ms    ±15.46%        9.54 ms       15.73 ms

Comparison: 
bench (pr)        138.85
bench              96.60 - 1.44x slower +3.15 ms

Memory usage statistics:

Name          Memory usage
bench (pr)        10.85 MB
bench             11.54 MB - 1.06x memory usage +0.69 MB

**All measurements for memory usage were the same**

##### With input single class #####
Name                 ips        average  deviation         median         99th %
bench (pr)        127.96        7.81 ms    ±14.07%        7.28 ms       11.16 ms
bench              88.56       11.29 ms    ±17.78%       10.79 ms       16.49 ms

Comparison: 
bench (pr)        127.96
bench              88.56 - 1.44x slower +3.48 ms

Memory usage statistics:

Name          Memory usage
bench (pr)        10.81 MB
bench             11.54 MB - 1.07x memory usage +0.74 MB

```

```
read_file = fn name ->
  __ENV__.file
  |> Path.dirname()
  |> Path.join(name)
  |> File.read!()
end

html_input = read_file.("medium.html")

Benchee.run(
  %{
    "bench" => fn selector -> Floki.Finder.find(html, selector)  end
  },
  time: 10,
  memory_time: 2,
  inputs: %{
    "single class" => ".mw-highlight",
    "multiple classes" => ".citation.pressrelease" ,
    "child" => ".citation > .external > *",
    "descendant" => ".external *" ,
  },
)
```

### DIFF
--- a/lib/floki/selector.ex
+++ b/lib/floki/selector.ex
@@ -144,7 +144,7 @@ defmodule Floki.Selector do
   defp do_classes_matches?(nil, _classes), do: false
 
   defp do_classes_matches?(class_attr_value, classes) do
-    classes -- String.split(class_attr_value, ~r/\s+/) == []
+    classes -- String.split(class_attr_value, [" ", "\t", "\n"]) == []
   end
 
   defp attributes_matches?(_node, []), do: true
--- a/lib/floki/selector.ex
+++ b/lib/floki/selector.ex
@@ -169,13 +169,15 @@ defmodule Floki.Selector do
   end
 
   defp do_classes_matches?(class_attr_value, [class]) do
-    Enum.member?(String.split(class_attr_value, [" ", "\t", "\n"]), class)
+    class_attr_value
+    |> String.split([" ", "\t", "\n"], trim: true)
+    |> Enum.member?(class)
   end
 
   defp do_classes_matches?(class_attr_value, classes) do
     min_size = Enum.reduce(classes, -1, fn item, acc -> acc + 1 + bit_size(item) end)
     can_match? = bit_size(class_attr_value) >= min_size
-    can_match? && classes -- String.split(class_attr_value, [" ", "\t", "\n"]) == []
+    can_match? && classes -- String.split(class_attr_value, [" ", "\t", "\n"], trim: true) == []
   end
 
   defp attributes_matches?(_node, []), do: true
--- a/lib/floki/selector.ex
+++ b/lib/floki/selector.ex
@@ -143,6 +143,11 @@ defmodule Floki.Selector do
 
   defp do_classes_matches?(nil, _classes), do: false
 
+  defp do_classes_matches?(class_attr_value, [class | _])
+       when bit_size(class_attr_value) < bit_size(class) do
+    false
+  end
+
   defp do_classes_matches?(class_attr_value, classes) do
     classes -- String.split(class_attr_value, [" ", "\t", "\n"]) == []
   end
--- a/lib/floki/selector.ex
+++ b/lib/floki/selector.ex
@@ -78,11 +78,26 @@ defmodule Floki.Selector do
   def match?(%Comment{}, _selector, _tree), do: false
 
   def match?(html_node, selector, tree) do
-    id_match?(html_node, selector.id) && namespace_match?(html_node, selector.namespace) &&
-      type_match?(html_node, selector.type) && classes_matches?(html_node, selector.classes) &&
+    can_match_combinator?(html_node, selector.combinator) &&
+      id_match?(html_node, selector.id) &&
+      namespace_match?(html_node, selector.namespace) &&
+      type_match?(html_node, selector.type) &&
+      classes_matches?(html_node, selector.classes) &&
       attributes_matches?(html_node, selector.attributes) &&
       pseudo_classes_match?(html_node, selector.pseudo_classes, tree)
   end
+
+  defp can_match_combinator?(_node, nil), do: true
+
+  defp can_match_combinator?(
+         %HTMLNode{children_nodes_ids: []},
+         %Selector.Combinator{match_type: match_type}
+       )
+       when match_type in [:child, :descendant] do
+    false
+  end
+
+  defp can_match_combinator?(_node, _combinator), do: true
 
   defp id_match?(_node, nil), do: true
   defp id_match?(%HTMLNode{attributes: []}, _), do: false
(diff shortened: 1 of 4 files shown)


lib/floki/selector.ex
```diff
@@ -143,8 +158,24 @@ defmodule Floki.Selector do
...
+    false
+  end
+
+  defp do_classes_matches?(class_attr_value, [class])
+       when bit_size(class_attr_value) == bit_size(class) do
+    class == class_attr_value
+  end
+
+  defp do_classes_matches?(class_attr_value, [class]) do
+    Enum.member?(String.split(class_attr_value, [" ", "\t", "\n"]), class)
```
philss, line 172: I think this is not exactly the same from the regex version, because the occurrence of multiple whitespace chars in sequence are treated differently.

For example, if we have a class attribute that contains multiple spaces between classes values, the former version would capture that correctly, while the new version does not:

```elixir
class_attr_value = "my class\tanother   one"

# new
String.split(class_attr_value, [" ", "\t", "\n"])
#=> ["my", "class", "another", "", "", "one"]

# old
String.split(class_attr_value, ~r/\s+/)
#=> ["my", "class", "another", "one"]
```

We can, however, reject the empty binaries from the match. Can you try this and see if there is an impact in performance?

lib/floki/selector.ex
```diff
@@ -143,8 +158,24 @@ defmodule Floki.Selector do
...
+  end
+
   defp do_classes_matches?(class_attr_value, classes) do
-    classes -- String.split(class_attr_value, ~r/\s+/) == []
+    min_size = Enum.reduce(classes, -1, fn item, acc -> acc + 1 + bit_size(item) end)
+    can_match? = bit_size(class_attr_value) >= min_size
+    can_match? && classes -- String.split(class_attr_value, [" ", "\t", "\n"]) == []
```
philss, line 178: Same thing here :)

lib/floki/selector/attribute_selector.ex
```diff
@@ -63,7 +63,7 @@ defmodule Floki.Selector.AttributeSelector do
     s.attribute
     |> get_value(attributes)
     # Splits by whitespaces ("a  b c" -> ["a", "b", "c"])
-    |> String.split(~r/\s+/)
+    |> String.split([" ", "\t", "\n"])
```
philss, line 66: Here as well. I think we need to consider the same problems here.

lib/floki/selector/attribute_selector.ex
```diff
@@ -103,7 +103,7 @@ defmodule Floki.Selector.AttributeSelector do
 
   def match?(attributes, s = %AttributeSelector{match_type: :includes, value: value}) do
     get_value(s.attribute, attributes)
-    |> String.split(~r/\s+/)
+    |> String.split([" ", "\t", "\n"])
```
philss, line 106: And here :)

lib/floki/selector/parser.ex
```diff
@@ -267,4 +267,8 @@ defmodule Floki.Selector.Parser do
     Logger.debug("Only simple selectors are allowed in :not() pseudo-class. Ignoring.")
     nil
   end
+
+  defp optimize_selector(selector) do
```
philss, line 271: I think it worth to add a comment here that we are doing this due to an improvement in the `Selector` algo. I think the description of the test is good enough.

> Reorders classes in selector to improve matching performance.

 WDYT?

philss (commented):
Nice optimizations! :rocket: 

Let's see if we can solve that issue with the whitespaces :)

lib/floki/selector.ex
ypconstante: Done, fa64b01261028d47eb5d41c1f6aab840a781f636.
Adding `trim: true` didn't had performance impact in the benchmark, but it did reduce the memory usage, since we won't have this empty strings in the list.

lib/floki/selector.ex
ypconstante: fa64b01261028d47eb5d41c1f6aab840a781f636

lib/floki/selector/attribute_selector.ex
ypconstante: fa64b01261028d47eb5d41c1f6aab840a781f636

lib/floki/selector/attribute_selector.ex
ypconstante: fa64b01261028d47eb5d41c1f6aab840a781f636

lib/floki/selector/parser.ex
ypconstante: Done e4a8a8a9c2c6106fe7c5c23543c39f2969a5b57f

philss (approved):

philss:
@ypconstante thank you!

Issue #515, opened 2023-12-26
Proposal: optimized find for simple cases


Valian:
## Feature goal

Currently, Floki.find always creates an HTMLTree, which seems like not always necessary. Consider finding all links, or elements with particular id or class. Basically all cases when we don't need to consider hierarchy of objects and it's enough to look at nodes one-by-one, with a simple traversal. 

In a project I'm involved in I noticed slowdowns in that area. A quick traversal is an order of magnitude faster, without memory allocation. 

Would you be interested in a PR covering such an optimization? I can try to work on this. 

PS Sorry for lack of examples, I'm writing from mobile 😅

philss:
@Valian I think it worth to explore that path, yes! I can't see a problem upfront. But like you said, this may work only for simple classes and ID selectors.

So please go ahead! Also, take a look at the latest merges - some of them were performance optimizations made by @ypconstante: https://github.com/philss/floki/commits/main/?author=ypconstante

Valian:
Thanks, I'll look into this! I know it might work only for simple checks like IDs, tag names and classes, but I have a feeling it might be the most common usage. Eg. finding all links is such an example, which I believe is extremely common. 

Will try to find some time on evenings in the next week - two to tackle this.

philss:
@ypconstante @Valian  do you think this can be closed?

Valian:
@philss Sorry recently my family got bigger 👦 and didn't really had time to work on this... But I see @ypconstante did a great job! Are your recent PRs enough to solve this or there is some work still needed?

ypconstante:
There are still some cases we can enable this optimization without too many changes, like the remaining combinators depending on the selector, and pseudo-classes that don't require the tree.

philss:
@ypconstante I'm closing this now. Once again, thank you for the great work you did to bring these performance gains! ❤️

PR #518, opened 2024-01-01
Use stack on Finder

ypconstante:
Today Finder uses Enum.reduce and nested calls to traverse_with to do it's filtering.

This works fine when using HTMLTree, but this structure makes it harder to optimize some cases in which the HTMLTree isn't necessary (https://github.com/philss/floki/issues/515).

This PRs updates Finder to use a stack, similar to what we have on HTMLTree.build_tree. This makes it easier to implement the change for the ticket above, or other future changes to operations that don't need the HTMLTree.build step.

I also did some other clean ups in this module, I've split each change in a separated commit.

Performance is basically the same, but when profiling the timing is a lot better and more readable because of the removal of Enum.reduce

Before - selector `span a * i` 
```
%                                                           CNT           ACC           OWN        
[{ "<0.99.0>",                                           278908,    undefined,      442.743}].   %%

{  {'Elixir.Enum','-reduce/3-lists^foldl/2-0-',3},        29692,      402.120,       50.704}.      
{  {'Elixir.Floki.Finder',traverse_with,4},               18660,      309.861,       50.072}.      
{  {'Elixir.Floki.Selector','match?',3},                  12986,      160.655,       42.984}.      
{  {'Elixir.Floki.HTMLTree',build_tree,5},                15867,       39.761,       28.100}.      
{  {'Elixir.String',split,3},                              6522,       73.182,       26.562}.      
```

After
```
%                                                           CNT           ACC           OWN        
[{ "<0.99.0>",                                           231183,    undefined,      363.851}].   %%

{  {'Elixir.Floki.Finder',traverse_with,4},               38959,      322.261,       66.273}.      
{  {'Elixir.Floki.Selector','match?',3},                  12986,      158.041,       42.423}.      
{  {'Elixir.Floki.HTMLTree',build_tree,5},                15867,       40.198,       28.321}.      
{  {'Elixir.String',split,3},                              6522,       71.817,       26.114}.       
```

### DIFF
--- a/lib/floki/finder.ex
+++ b/lib/floki/finder.ex
@@ -28,17 +28,6 @@ defmodule Floki.Finder do
     find_selectors(html_tree, selectors)
   end
 
-  @spec map(Floki.html_tree() | Floki.html_node(), function()) ::
-          Floki.html_tree() | Floki.html_node()
-
-  def map({name, attrs, rest}, fun) do
-    {new_name, new_attrs} = fun.({name, attrs})
-
-    {new_name, new_attrs, Enum.map(rest, &map(&1, fun))}
-  end
-
-  def map(other, _fun), do: other
-
   defp find_selectors(html_tuple_or_list, selectors) do
     tree = HTMLTree.build(html_tuple_or_list)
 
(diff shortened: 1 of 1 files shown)


philss (approved):
Makes totally sense! Really cool! :shipit:

PR #522, opened 2024-01-10
Polymorphism on Finder.find

ypconstante:
Today Finder.find expects a list of html_node tuples, and returns an HTMLTree and a list of resulting HTMLNode.

To allow the changes for #515, we need to return html_nodes only when we don't need to later use the HTMLTree. If receiving an HTMLTree, Finder.find will return a  list of HTMLNode, and if receiving a list of html_nodes the return will be a list of html_nodes. 

For now we'll keep building an HTMLTree for the search, but this allow a future change to avoid building the HTMLTree on simple selectors without having to change modules using Finder.

This is a breaking change for projects using Finder.find directly.

### DIFF
--- a/lib/floki.ex
+++ b/lib/floki.ex
@@ -280,17 +280,13 @@ defmodule Floki do
     )
 
     with {:ok, document} <- Floki.parse_document(html) do
-      {tree, results} = Finder.find(document, selector)
-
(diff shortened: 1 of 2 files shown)


philss:
Hey @ypconstante , thanks for the PR once again! I will hold this for a bit, because I need to release a new version - a patch one. I know `Finder` is not public API, but just in case I will keep the out of the next version.

I will be back to review it soon :)

PR #523, opened 2024-01-10
Run all selector test with tuple list and html tree

ypconstante:
To make sure we don't have any regression on the changes for https://github.com/philss/floki/issues/515, we need to run all selector tests with both html node tuple list and with HTMLTree.

This PR modifies most asserts on Floki.find to use `assert_find`, which will check the results of Floki.find with an html node tuple list, and then check if the result is the same when calling Floki.Finder.find with an HTMLTree. 

On tests that have to use pattern matching instead of equals, the comparison is done directly in the test instead of using `assert_find`.

Depends on #522

### DIFF
--- a/lib/floki.ex
+++ b/lib/floki.ex
@@ -280,17 +280,13 @@ defmodule Floki do
     )
 
     with {:ok, document} <- Floki.parse_document(html) do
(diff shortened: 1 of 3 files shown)


philss (approved):
Makes sense! Nice :)

PR #533, opened 2024-02-09
Optimize Floki.children

ypconstante:
Floki.children is called a lot by Phoenix LiveView when doing tests. On the project I'm working on, this functions gets called more than 8 million times when running all of our tests. 

This PR adds some pattern matching to this function to avoid using `Access.get` if possible.

```
Name                      ips        average  deviation         median         99th %
children (pr)         11.50 K       86.95 μs   ±239.29%       53.26 μs     1559.21 μs
children (main)        7.19 K      139.07 μs   ±145.06%      104.25 μs     1507.63 μs

Comparison: 
children (pr)         11.50 K
children (main)        7.19 K - 1.60x slower +52.12 μs
``` 
```

read_file = fn name ->
  __ENV__.file
  |> Path.dirname()
  |> Path.join(name)
  |> File.read!()
  |> Floki.parse_document!()
end

doc = read_file.("small.html")
html_nodes = Floki.find(doc, "*")

Benchee.run(
  %{
    "children" => fn ->
      for html_node <- html_nodes do
        Floki.children(html_node)
        Floki.children(html_node, include_text: false)
        Floki.children(html_node, include_text: true)
      end
    end
  },
  time: 10,
  save: [path: "benchs/results/finder-#{tag}.benchee", tag: tag],
  memory_time: 2
)
```

### DIFF
--- a/test/floki_test.exs
+++ b/test/floki_test.exs
@@ -1564,10 +1564,17 @@ defmodule FlokiTest do
   # Floki.children/2
 
   test "returns the children elements of an element including the text" do
-    assert Floki.children({"div", [], ["a parent", {"span", [], ["a child"]}]}) == [
-             "a parent",
-             {"span", [], ["a child"]}
-           ]
+    html_node = {"div", [], ["a parent", {"span", [], ["a child"]}]}
+
+    expected = [
+      "a parent",
+      {"span", [], ["a child"]}
+    ]
+
+    assert Floki.children(html_node) == expected
+    assert Floki.children(html_node, include_text: true) == expected
+    assert Floki.children(html_node, include_text: true, unknown_option: true) == expected
+    assert Floki.children(html_node, unknown_option: true) == expected
   end
 
   test "returns the children elements of an element without the text" do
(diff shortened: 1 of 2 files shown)


test/floki_test.exs
```diff
@@ -1564,10 +1564,17 @@ defmodule FlokiTest do
...
+    html_node = {"div", [], ["a parent", {"span", [], ["a child"]}]}
+
+    expected = [
+      "a parent",
+      {"span", [], ["a child"]}
+    ]
+
+    assert Floki.children(html_node) == expected
+    assert Floki.children(html_node, include_text: true) == expected
+    assert Floki.children(html_node, include_text: true, unknown_option: true) == expected
```
philss, line 1576: Why "unknown_option" is needed here?

test/floki_test.exs
```diff
@@ -1576,14 +1583,19 @@ defmodule FlokiTest do
...
-             {"span", [], ["child 1"]},
-             {"span", [], ["child 2"]}
-           ] = Floki.children(elements, include_text: false)
+    expected = [
+      {"span", [], ["child 1"]},
+      {"span", [], ["child 2"]}
+    ]
+
+    assert Floki.children(elements, include_text: false) == expected
+    assert Floki.children(elements, include_text: false, unknown_option: true) == expected
```
philss, line 1592: Same question here. I think we could filter out unknown options - maybe with `Keyword.validate!/2`, but we would need to remove support for Elixir 1.12. WDYT?

ypconstante: This is to have coverage in the fallback when we are not able to pattern match the include_text value directly
```
  def children({_, _, _} = html_node, opts) do
    children(html_node, include_text: opts[:include_text])
  end
```
Not sure if there's a better way to cause this to fail besides `Keyword.validate!/2`, I didn't even knew about this function

philss: OK, I think I will try to refactor that once I drop support for Elixir 1.12. Thanks!

PR #540, opened 2024-02-17
Always return find elements in the correct order

### LINKED ISSUE #539: Unexpected order using `find` with comma separated selectors in v0.35.3
Upgrading from 0.35.2 to 0.35.3, `find` no longer conforms with the order of the document elements for comma separated selectors.

This is what 0.35.2 produces:

```elixir
iex(1)> Floki.find("<div class=\"first\"></div><p class=\"second\"></p><div class=\"third\"></div>", "p,div")

[
  {"div", [{"class", "first"}], []},
  {"p", [{"class", "second"}], []},
  {"div", [{"class", "third"}], []}
]

```

This is what 0.35.3 produces:

```elixir
iex(1)> Floki.find("<div class=\"first\"></div><p class=\"second\"></p><div class=\"third\"></div>", "p,div")

[
  {"p", [{"class", "second"}], []},
  {"div", [{"class", "first"}], []},
  {"div", [{"class", "third"}], []}
]
```

ypconstante:
Instead of expecting `traverse_html_tree` to return elements in the correct order, we'll now use the `node_id` to sort find results.
Fixes #539

### DIFF
--- a/lib/floki/finder.ex
+++ b/lib/floki/finder.ex
@@ -47,7 +47,7 @@ defmodule Floki.Finder do
     stack = Enum.map(selectors, fn s -> {s, node_ids} end)
 
     traverse_html_tree(stack, tree, [])
-    |> Enum.reverse()
+    |> Enum.sort_by(& &1.node_id)
     |> Enum.uniq()
   end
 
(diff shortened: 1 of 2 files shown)


philss:
@ypconstante thank you!

PR #541, opened 2024-02-18
Update Elixir version requirement to 1.13

vittoriabitton:


philss (commented):
@vittoriabitton Thank you! :D

Please change the workflow file to test against 1.13 instead of 1.12.

And also: I'm going to release a new patch version tomorrow, before merging this one.

PR #547, opened 2024-02-28
Implement Inspect protocol for HTMLTree

vittoriabitton:


lib/floki/html_tree.ex
```diff
@@ -328,4 +328,78 @@ defmodule Floki.HTMLTree do
...
+    import Inspect.Algebra
+
+    def inspect(html_tree, opts) do
+      open = "%HTMLTree{"
+      close = "}"
+      container_opts = [separator: "", break: :flex]
+
+      container_doc(
+        open,
+        [{html_tree, html_tree.root_nodes_ids}],
```
philss, line 342: I think we could get the nodes and pass it right here, instead of traversing inside `fun/2`. It would carry the `html_tree` only for nodes that are of `HTMLNode` type.

Something like this:

```suggestion
        nodes_maybe_with_tree(html_tree, html_tree.root_nodes_ids),
```

And the function definition would be something like this:

```elixir
    defp nodes_maybe_with_tree(html_tree, nodes_ids) do
      nodes_ids
      |> Enum.reverse()
      |> Enum.map(fn node_id ->
        case Map.get(html_tree.nodes, node_id) do
          %HTMLNode{} = html_node ->
            {html_tree, html_node}

          other ->
            other
        end
      end)
    end
```
PS: It could be a `with` instead of a `case`.

lib/floki/html_tree.ex
```diff
@@ -328,4 +328,78 @@ defmodule Floki.HTMLTree do
...
+      end)
+      |> concat()
+    end
+
+    defp build_tree(%Comment{content: comment}, _opts),
+      do: {comment, "", [separator: "", break: :strict]}
+
+    defp build_tree(%Text{content: text}, _opts), do: {text, "", [separator: "", break: :strict]}
+
+    defp build_tree(%HTMLNode{} = node, opts) do
```
philss, line 379: With that suggestion, you could simplify the `fun/2` implementation.

```suggestion
    defp fun({html_tree, %HTMLNode{} = html_node}, opts) do
      {open, close, container_opts} = build_node(html_node, opts)

      container_doc(
        open,
        nodes_maybe_with_tree(html_tree, html_node.children_nodes_ids),
        close,
        opts,
        &fun/2,
        container_opts
      )
    end

    defp fun(%Comment{content: comment}, opts),
      do: color(concat(["<!-- ", comment, " -->"]), :comment, opts)

    defp fun(%Text{content: text}, opts), do: color(text, :string, opts)

    defp build_node(%HTMLNode{} = node, opts) do
```

test/floki/html_tree_test.exs
```diff
@@ -256,4 +256,42 @@ defmodule Floki.HTMLTreeTest do
...
+
+  test "Inspect works with HTMLTree" do
+    html_tree = %HTMLTree{
+      root_nodes_ids: [2, 1],
+      node_ids: [2, 1],
+      nodes: %{
+        1 => %Text{content: "hello", node_id: 1},
+        2 => %Text{content: " world", node_id: 2}
+      }
+    }
```
philss, line 268: After thinking for a while, I believe the parser won't produce such tree, because the `Text` would be collapsed`. So I would change this to one `Text` node, or maybe adding a tag in between.

test/floki/html_tree_test.exs
```diff
@@ -256,4 +256,42 @@ defmodule Floki.HTMLTreeTest do
...
+          node_id: 3
+        },
+        4 => %HTMLNode{type: "b", parent_node_id: 3, children_nodes_ids: [5], node_id: 4},
+        5 => %Text{content: "click me", parent_node_id: 4, node_id: 5},
+        6 => %HTMLNode{type: "span", parent_node_id: 1, node_id: 6}
+      }
+    }
+
+    assert inspect(html_tree) ==
+             ~s|%HTMLTree{<html>start of the stack<a class="link"><b>click me</b></a><span></span></html>}|
```
philss, line 295: We forgot to format the comment. The suggestions I made is changing that. But the output here will be different.

PS: try to add a `|> dbg()` to line 292 and see the beauty of the output :D

philss (commented):
This is looking great! 🎉 

I added some refactoring suggestions :)

test/floki/html_tree_test.exs
```diff
@@ -256,4 +256,42 @@ defmodule Floki.HTMLTreeTest do
...
+      nodes: %{
+        1 => %Text{content: "hello world", node_id: 1}
+      }
+    }
+
+    assert inspect(html_tree) ==
+             ~s|#Floki.HTMLTree[hello world]|
+  end
+
+  test "Inspect works with HTMLTree with nested nodes" do
```
philss, line 273:
```suggestion
  test "inspect works with HTMLTree with nested nodes" do
```

test/floki/html_tree_test.exs
```diff
@@ -256,4 +256,42 @@ defmodule Floki.HTMLTreeTest do
     refute Enum.member?(html_tree, %{html_node3 | type: "marquee"})
     refute Enum.member?(html_tree, 42)
   end
+
+  test "Inspect works with HTMLTree" do
```
philss, line 260:
```suggestion
  test "inspect works with HTMLTree" do
```

philss (approved):
Awesome!! :D
Thank you! 💜

PR #548, opened 2024-03-01
add Floki.find_by_id/2

SteffenDE:
Related to the problem and discussion in https://github.com/phoenixframework/phoenix_live_view/issues/3144.

In LiveViewTest we need to find elements by their id. Currently, we use `Floki.find(html, "##{id}")`, but this breaks as soon as there are special characters in the id. We could use `Floki.find(html, %Floki.Selector{id: id})`, but the `Floki.Selector` struct is not documented and therefore I'm not sure if it's something to rely on.

As `document.getElementById` is also a frequently used API in the browser, adding a `find_by_id` or similar function to Floki could be quite useful.

This PR adds a simple find_by_id function that internally uses the `%Floki.Selector{id: id}` struct.
As using find with a string is deprecated, I did not add this functionality to find_by_id.

### DIFF
--- a/lib/floki.ex
+++ b/lib/floki.ex
@@ -290,6 +290,32 @@ defmodule Floki do
   end
 
   @doc """
+  Finds the first element in an HTML tree by id.
+
+  Returns `nil` if no element is found.
+
+  This is useful when there are IDs that contain special characters that
+  are invalid when passed as is as a CSS selector.
+  It is similar to the `getElementById` method in the browser.
+
+  ## Examples
+
+      iex> {:ok, html} = Floki.parse_fragment(~s[<p><span class="hint" id="id?foo_special:chars">hello</span></p>])
+      iex> Floki.get_by_id(html, "id?foo_special:chars")
+      {"span", [{"class", "hint"}, {"id", "id?foo_special:chars"}], ["hello"]}
+      iex> Floki.get_by_id(html, "does-not-exist")
+      nil
+
+  """
+  @spec get_by_id(html_tree() | html_node(), String.t()) :: html_tree
+  def get_by_id(html_tree_as_tuple, id)
+      when is_list(html_tree_as_tuple) or is_html_node(html_tree_as_tuple) do
+    html_tree_as_tuple
+    |> Finder.find(%Floki.Selector{id: id})
+    |> List.first()
+  end
+
+  @doc """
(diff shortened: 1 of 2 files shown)


SteffenDE:
Maybe we should also only return nil or the first element found? That would mirror the behavior of `document.getElementById`. Not sure what Floki does when ids are not unique. We could also raise if there are multiple results.

philss:
> Maybe we should also only return nil or the first element found?

@SteffenDE yeah, I agree. We don't need to return a list here.

> The function could be called find_one_by_id or find_first_by_id to make it more explicit.

I believe if we document well, it's OK to use `find_by_id`. WDYT @josevalim ?

lib/floki.ex
```diff
@@ -289,6 +289,33 @@ defmodule Floki do
     Finder.find(html_tree_as_tuple, selector)
   end
 
+  @doc """
+  Finds the first element in an HTML tree by id.
+
+  Returns nil if no element is found.
```
philss, line 295:
```suggestion
  Returns `nil` if no element is found.
```

lib/floki.ex
```diff
@@ -289,6 +289,33 @@ defmodule Floki do
...
+      nil
+
+  """
+  @spec find_by_id(html_tree() | html_node(), String.t()) :: html_tree
+  def find_by_id(html_tree_as_tuple, id)
+      when is_list(html_tree_as_tuple) or is_html_node(html_tree_as_tuple) do
+    case Finder.find(html_tree_as_tuple, %Floki.Selector{id: id}) do
+      [] -> nil
+      [node | _] -> node
+    end
```
philss, line 316: What about using `List.first/1` here?

```suggestion
    html_tree_as_tuple
    |> Finder.find(%Floki.Selector{id: id})
    |> List.first()
```

philss (commented):
Only minor points, but LGTM!

josevalim:
What about calling it “get” to mirror browsers then? Or get_by_id?

philss:
> What about calling it “get” to mirror browsers then? Or get_by_id?

I like the idea of using `get_by_id`! WDYT @SteffenDE ?

philss (approved):
🚢

philss:
@SteffenDE thank you for the contribution! 

and thanks @ypconstante and @josevalim for the points and the discussion!

PR #549, opened 2024-03-01
fix typespec of get_by_id/2

SteffenDE:
well... so I forgot to change the typespec when changing the function to return a node or nil instead of a tree.

This fixes that.


philss:
oh, I didn't see that either 😬 
Thanks!

PR #555, opened 2024-03-13
raw_html treats the content of title tags as plain text

aymanosman:
`parse_document` correctly treats the content of title tags as plain text. But `raw_html` does not. This PR fixes that.


aymanosman:
@philss I changed the test to avoid the differences between the parsers. I've ran `mix test`, `mix test_fast_html` and `mix test_html5ever` locally to verify all the tests pass.

philss:
@aymanosman thank you! And sorry for the delay!

Issue #552, opened 2024-03-05
Floki.raw_html/2 no longer respects `config :floki, :encode_raw_html, false`


JohnnyCurran:
## Description

A clear and concise description of what the bug is, including the code used
and the exception stacktrace, if any.

## To Reproduce

- Using Floki v0.36.0
- Using Elixir v1.16.0
- Using Erlang OTP v26.2.1

## Expected behavior

`Phoenix.LiveViewTest.render/1` wraps `Floki.raw_html/2` in order to process the html. We use assertions for things that contain apostrophes so we don't have to encode them:

```elixir
assert html =~ "I'm an apostrophe"
```

so that we don't have to write:

```elixir
assert html =~ "I&#39;m an apostrophe"
```

On 0.35.2, we added `config :floki, :encode_raw_html, false` to `config/config.exs` and that solved our problem. But it appears that special characters are always HTML encoded by default now

PR #561, opened 2024-04-19
fix: read encode_raw_html config as default for raw_html encode option

Sgoettschkes:
As discussed in #552, with the recent changes to how the opts are read in the RawHTML module, the `encode_raw_html` configuration option is never read.

With this change, the `encode_raw_html` value set in the environment is used as the default or true if it does not exist, which is in line with how it behaved up until now.


philss:
@Sgoettschkes Makes sense! Thank you! 💜

PR #587, opened 2024-07-29
Copy and paste the benchmark commands

kianmeng:


kianmeng:
:partying_face: :partying_face: :partying_face: :partying_face: :partying_face:

PR #603, opened 2024-12-05
Fix bug propagating identity encoder in `raw_html/2`

andyleclair:
Hi! We noticed a bug when going to upgrade Floki. It seems that any tag that is set to the `identity` encoder would propagate that encoder through the rest of the tree (instead of just to its' children). I've added a test case here taken from some real Appcues content that triggers this behavior. Without this fix, it's _very_ possible for Floki to emit broken HTML. If you parse any `script` or `style`, tag, the identity encoder would be applied, which would in this case, emit an un-escaped `data-attrs-event` tag (even if escaping is explicitly enabled!). I'm dubious on if it's the duty to handle not decoding the entities in the parser, but I'm 100% sure this is a bug as-is.

### DIFF
--- a/lib/floki/raw_html.ex
+++ b/lib/floki/raw_html.ex
@@ -133,7 +133,7 @@ defmodule Floki.RawHTML do
          self_closing_tags,
          line_ending
        ) do
-    encoder =
+    curr_encoder =
       case type do
         "script" -> @no_encoder
         "style" -> @no_encoder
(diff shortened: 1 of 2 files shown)


lib/floki/raw_html.ex
```diff
@@ -133,7 +133,7 @@ defmodule Floki.RawHTML do
          self_closing_tags,
          line_ending
        ) do
-    encoder =
+    curr_encoder =
```
ypconstante, line 136: There's another bug here, the identity encoder is also being applied to the script/style/title attributes.
This encoder change should be done inside of `case children` to avoid issues, and will also fix the bug you found.

andyleclair: Ah you're right! I've updated the test and moved that check

philss:
@andyleclair good call! Thank you!

andyleclair:
Thank you @philss! I'd love it if you didn't mind cutting a release with this fix in it.

philss:
@andyleclair sure! I did a minor version release: `v0.37.0`. LMK if this works for you.

Issue #604, opened 2024-12-10
Crash at (floki 0.37.0) lib/floki/finder.ex:56: Floki.Finder.find/2


mellelieuwes:
## Description

Error in Phoenix Live View test code. The error:

16:43:41.594 [error] GenServer #PID<0.1169.0> terminating
** (MatchError) no match of right hand side value: []
    (floki 0.37.0) lib/floki/finder.ex:56: Floki.Finder.find/2
    (phoenix_live_view 1.0.0) lib/phoenix_live_view/test/dom.ex:70: Phoenix.LiveViewTest.DOM.targets_from_selector/2
    (phoenix_live_view 1.0.0) lib/phoenix_live_view/test/client_proxy.ex:1063: anonymous fn/6 in Phoenix.LiveViewTest.ClientProxy.maybe_js_commands/6
    (elixir 1.17.0) lib/enum.ex:4353: Enum.flat_map_list/2
    (phoenix_live_view 1.0.0) lib/phoenix_live_view/test/client_proxy.ex:321: Phoenix.LiveViewTest.ClientProxy.handle_info/2
    (stdlib 6.1.2) gen_server.erl:2345: :gen_server.try_handle_info/3
    (stdlib 6.1.2) gen_server.erl:2433: :gen_server.handle_msg/6
    (stdlib 6.1.2) proc_lib.erl:329: :proc_lib.init_p_do_apply/3
[...]

PR #605, opened 2024-12-10
Prevents Floki.Finder.find/2 from crashing

mellelieuwes:
See issue #604


philss:
Makes sense to me. What do you think, @ypconstante ?

philss:
BTW, thank you for the report and the fix, @mellelieuwes ! <2

ypconstante:
Since the function is recursive, this will probably cause us to use the expensive traversal in some cases in which we could use the fast one. I think we should handle this in the `find` function directly.
Is this error triggered in any case besides an empty selector `""`?

Issue #592, opened 2024-10-01
Selecting parents


glyh:
## Feature goal

When I encounter some webpage like this:
```html
<x><y>hey</y> wat</x>
```

I want to select the tag x, and get the text " wat". The characteristic for this tag is that it's a *parent* of some other tag containing text "hey". 

There doesn't seem to be one help me to do so. For now I just use "E:nth-of-type" to  find it by index, but this seems fragile as once the element moves this will break. 

## Dependencies
None

philss:
As @bvobart mentioned, we now have the `:has` pseudo-selector (from v0.38). I think you can work with a combination of `:has(y:fl-contains('hey'))` and `Floki.text()` with the `:deep` option set to `false`. See the docs: https://hexdocs.pm/floki/Floki.html#text/2

I'm going to close this one, but feel free to comment if you have any questions.

Issue #382, opened 2022-02-02
Extend `fl-contains` to mach on children text as well


revati:
## Feature goal

Extend `fl-contains` to mach on children text as well.

```elixir
"<div>Some</div>"
|> Floki.parse_document!()
|> Floki.find("div:fl-contains('Some')")
|> IO.inspect(label: :first)
# Output: first: [{"div", [], ["Some"]}]

"<div><span>Some</span></div>"
|> Floki.parse_document!()
|> Floki.find("div:fl-contains('Some')")
|> IO.inspect(label: :second)
# Output: second: []
# Desired output: [{"div", [], [{"span", [], ["Some"]}]}]
```

philss:
Hey @revati :wave: 

You actually can archive similar results using a _descendant combinator_ with the "all" selector ("`*`").
Please try:

```elixir
"<div><span>Some</span></div>"
|> Floki.parse_document!()
|> Floki.find("div *:fl-contains('Some')")
|> IO.inspect(label: :second)
# Output: second: [{"span", [], ["Some"]}]
# Desired output: [{"div", [], [{"span", [], ["Some"]}]}]
```

bvobart:
@philss I'm having the same issue. Thanks for the suggestion, but there could be other interesting items in the `div` that you might want to fetch. Given that there's currently no way to select the parent of the span you find (see #592 and #482), there's no way to fetch that `div`.

Consider also this example, a mini version of my current use case:

```html
"""
<table>
  <tbody>
    <tr>
      <th><label>TEST</label></th>
      <td>fetch me pls</td>
    </tr>
  </tbody>
</table>
"""
|> Floki.parse_fragment!()
|> Floki.find("th:fl-contains('TEST') ~ td")
```

My intention is to fetch the `td` element following the `th` containing the text TEST. However, this currently does not work (the output is simply `[]`), as `th:fl-contains('TEST')` does not match `<th><label>TEST</label></th>` due to the text being wrapped with a `label` element. An obvious solution would be to allow `fl-contains` to support matching on the text in child elements as well, which would make my query work unchanged. However, I understand that that changes the semantics of `fl-contains` significantly, likely _so_ much that it's a breaking change. 

So another solution that would solve this issue, is to have some form of parent selector, or more specifically, to implement `:has` (#482). This would solve the issue by allowing the following query:

```
th:has(*:fl-contains('TEST')) ~ td
```

[...]

PR #624, opened 2025-05-15
feat: implement :has pseudo-selector functionality

### LINKED ISSUE #482: Support for :has pseudo selector
## Feature goal

(From JSoup [docs](https://jsoup.org/cookbook/extracting-data/selector-syntax)) 

> :has(selector): find elements that contain elements matching the selector; e.g. div:has(p)

```elixir
input = “<div><p>foo</p><h2>$100</h2></div><div><h1>foo</h1></div>”

result = Floki.find(html, “div:has(p) > h2”)
```

bvobart:
Fixes #482, #382, #592  

TODO:
- [x] Add documentation for `:has` to README.
- [x] Review by @philss
- [x] Fix tests
- [x] Add a few more tests to include other combinators and sub-selectors.

### DIFF
--- a/lib/floki/selector/pseudo_class.ex
+++ b/lib/floki/selector/pseudo_class.ex
@@ -153,8 +153,8 @@ defmodule Floki.Selector.PseudoClass do
   end
 
   def match_has?(tree, html_node, %__MODULE__{value: value} = pseudo_class) do
-    Enum.all?(value, fn inner_selector ->
-     Enum.any?(html_node.children_nodes_ids, fn id ->
+    Enum.any?(value, fn inner_selector ->
+      Enum.any?(html_node.children_nodes_ids, fn id ->
         child = Map.fetch!(tree.nodes, id)
 
         is_struct(child, HTMLNode) and
--- a/lib/floki/selector/pseudo_class.ex
+++ b/lib/floki/selector/pseudo_class.ex
@@ -152,6 +152,20 @@ defmodule Floki.Selector.PseudoClass do
     html_node.node_id in tree.root_nodes_ids
   end
 
+  def match_has?(tree, html_node, %__MODULE__{value: value} = pseudo_class) do
+    value
+    |> Enum.all?(fn inner_selector ->
+      html_node.children_nodes_ids
+      |> Enum.any?(fn id ->
+        child = Map.get(tree.nodes, id)
+
+        child |> is_struct(HTMLNode) and
+          (child |> Floki.Selector.match?(inner_selector, tree) or
+             match_has?(tree, child, pseudo_class))
+      end)
+    end)
+  end
+
   defp node_position(ids, %HTMLNode{node_id: node_id}) do
     position = Enum.find_index(ids, fn id -> id == node_id end)
     position + 1
--- a/lib/floki/selector/pseudo_class.ex
+++ b/lib/floki/selector/pseudo_class.ex
@@ -153,14 +153,12 @@ defmodule Floki.Selector.PseudoClass do
   end
 
   def match_has?(tree, html_node, %__MODULE__{value: value} = pseudo_class) do
-    value
-    |> Enum.all?(fn inner_selector ->
-      html_node.children_nodes_ids
-      |> Enum.any?(fn id ->
-        child = Map.get(tree.nodes, id)
+    Enum.all?(value, fn inner_selector ->
+     Enum.any?(html_node.children_nodes_ids, fn id ->
+        child = Map.fetch!(tree.nodes, id)
 
-        child |> is_struct(HTMLNode) and
-          (child |> Floki.Selector.match?(inner_selector, tree) or
+        is_struct(child, HTMLNode) and
+          (Floki.Selector.match?(child, inner_selector, tree) or
              match_has?(tree, child, pseudo_class))
       end)
     end)
--- a/test/floki_test.exs
+++ b/test/floki_test.exs
@@ -1,4 +1,5 @@
 defmodule FlokiTest do
+  @moduledoc false
   use ExUnit.Case, async: true
 
   doctest Floki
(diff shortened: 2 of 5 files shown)


test/floki_test.exs
```diff
@@ -1577,6 +1578,305 @@ defmodule FlokiTest do
...
+    #   {"th", [], [{"div", [], [{"label", [], ["NESTED"]}]}]}
+    # ])
+
+    ## NOTE: this parses incorrectly, parses as:
+    ##  %PseudoClass{name: "not", value: [%Selector{type: "label", pseudo_classes: [%PseudoClass{name: "has", value: []}]}]}
+    ## but would expect to parse as:
+    ##  %PseudoClass{name: "not", value: [%Selector{type: "*", pseudo_classes: [%PseudoClass{name: "has", value: [%Selector{type: "label"}]}]}]}
+    # assert_find(html, "tr:not(:has(label))", [
+    #   {"tr", [], [{"th", [], ["No Label"]}, {"td", [], ["some data"]}]}
+    # ])
```
bvobart, line 1819: @philss During testing, I found some cases that are valid according to the CSS spec, but are not handled correctly by Floki. I suspect this has something to do with only simple selectors being allowed in `not` and `has`.

Perhaps out of scope for this PR, but do you see a solution for this? What would it take to support full selectors in `not` and `has`, do you think?

philss: Yeah, this is due the fact that we don't parse the selector correctly. I think we would need to tweak the "selector" parser to be more recursive when encounter those two pseudo-selectors. We would probably need a special case to handle the `>` combinator as well.

lib/floki/selector/pseudo_class.ex
```diff
@@ -152,6 +152,20 @@ defmodule Floki.Selector.PseudoClass do
...
+    |> Enum.all?(fn inner_selector ->
+      html_node.children_nodes_ids
+      |> Enum.any?(fn id ->
+        child = Map.get(tree.nodes, id)
+
+        child |> is_struct(HTMLNode) and
+          (child |> Floki.Selector.match?(inner_selector, tree) or
+             match_has?(tree, child, pseudo_class))
+      end)
+    end)
```
philss, line 166: Just some style preferences:

```suggestion
    Enum.all?(value, fn inner_selector ->
     Enum.any?(html_node.children_nodes_ids, fn id ->
        child = Map.fetch!(tree.nodes, id)

        is_struct(child, HTMLNode) and
          (Floki.Selector.match?(child, inner_selector, tree) or
             match_has?(tree, child, pseudo_class))
      end)
    end)
```

1. the pipe operator is nice, but I prefer to avoid it when there is only one step to pipe.
2. we expect that the node is always present in the tree, so using `Map.fetch!/2` is more assertive.

philss: I'm curious why we need to filter only for `HTMLNode` here. Is there any problem with matching against all nodes?

test/floki_test.exs
```diff
@@ -1577,6 +1578,314 @@ defmodule FlokiTest do
     ])
   end
 
+  test "has pseudo-class simple" do
+    html =
```
philss, line 1582: Most of the failures are due to the differences in the tree generated by the parsers. Please try to generate a tree that would be parsed the same, or add a tag to this test case with `@tag only_parser: Mochiweb`.

philss (commented):
Thank you! I think you nailed it!
I added some points to change, but overall it looks great!

lib/floki/selector/pseudo_class.ex
bvobart: > I'm curious why we need to filter only for HTMLNode here.

Because the selectors inside `:has()` can only match on HTML elements (`HTMLNode`), not on `Text` or `Comment` nodes.

test/floki_test.exs
bvobart: Ahh I didn't realise that the tests would be run against multiple parsers. Will fix!

test/floki_test.exs
bvobart: Yeah the selector parser would definitely need to be more recursive. I'm not familiar with the syntax of the `.xrl` files and accompanying parser implementations, but I did find compiler construction very interesting in uni, so here's some pseudo-grammar of how I think a CSS selector parser is supposed to be defined:

```elixir
# Let's say an identifier is just a word
ID = ~r/\w+/
# a number is just, well, a number
NUMBER = ~r/\d+/
# and a value is anything within quotes
VALUE = ~r/".*"/

# Then the grammar for Floki's CSS selectors can be defined recursively as:
SELECTOR = (
  :root
  *
  ID
  .ID
  SELECTOR[ID=VALUE]
  SELECTOR[ID~=VALUE]
  SELECTOR[ID^=VALUE]
  SELECTOR.nth-child(NUMBER)
  SELECTOR#ID
  # etc
  SELECTOR? SELECTOR
  SELECTOR? > SELECTOR
  SELECTOR? + SELECTOR
  SELECTOR? ~ SELECTOR
  SELECTOR?:checked
  SELECTOR?:disabled
  SELECTOR?:fl-contains(VALUE)
  SELECTOR?:fl-icontains(VALUE)
  SELECTOR?:not(SELECTOR)
  SELECTOR?:has(SELECTOR)
)
```

Note that this grammar allows a superset of what CSS selectors semantically allow, as for example `div:has(:has(img))` is not allowed in CSS, but `div:not(:has(:not(img)))` is. But as you can see, CSS selectors are quite recursive in their grammar, so the parser would need to be as well. Semantic correctness should be checked separately, potentially even at evaluation time.

Regarding selectors like `> h2`, whether on its own or within `:has`, e.g. `div:has(> h2)`, you could interpret them as the case `SELECTOR? > SELECTOR`, where the first selector is nil. Logically, the absence of a selector implies no selection is being done, i.e. every element we try, matches. Given that we start the evaluation of a CSS selector by trying all HTML nodes in the HTML tree, `> h2` on its own is thus equivalent to `* > h2` (i.e. match every HTML element that has a `h2` as a direct child). However, within `div:has(> h2)`, when we reach `> h2` in our evaluation, we have already narrowed down our search to match only `div`s, so `> h2` is only evaluated on all `div` elements in the tree, making `div:has(> h2)` essentially equivalent to `div > h2`. The same logic can be applied for other combinators: `div:has(+ dt)` is essentially the same as `div + dt`. That is how I interpret and logically make sense of the [`:has` CSS spec](https://www.w3.org/TR/selectors-4/#has-pseudo).

Does this help? I realise it's been a pretty long response and it's pretty difficult to get on the same page in a complex topic like this without speaking a common grammar, so if you want to have a video call someday to meet each other and have a more interactive conversation on this subject, I'd be happy to 😊

test/floki_test.exs
bvobart: Fixed! All tests should pass now :)

bvobart:
I noticed I made a mistake in my implementation. According to https://developer.mozilla.org/en-US/docs/Web/CSS/:has#logical_operations when we give `:has` multiple arguments, e.g. `:has(s1, s2)`, it means that `s1` OR `s2` must match in order for the `:has` to match. Initially, I thought it was an AND, which also made `:has(s1, s2)` equivalent to `:has(s1):has(s2)`, but that is not true.

I've fixed that now and the updated tests reflect that :)

test/floki_test.exs
philss: Yeah, that was super helpful! I will not promise to implement changes for that soon, but I will keep all of that in mind.
Thank you very much for your research and suggestions.

philss (approved):

PR #627, opened 2025-06-02
Add `:style` flag to text/2` docs

s3cur3:
This adds documentation for the `:style` keyword argument, which was previously demonstrated in the code examples, but not listed explicitly in the options section.


philss:
@s3cur3 thank you!

PR #632, opened 2025-06-17
Fix build status badge

sgerrand:
💁 GitHub, in their infinite wisdom, changed the format of the URL used for status badges without making them backwards compatible. This change fixes the image URL format and updates the hyperlink to the CI workflow.


philss:
@sgerrand thank you!

PR #634, opened 2025-06-23
Update v0.38.0 changelog item for breaking changes

kianmeng:
While upgrading deps from 0.37.1, we've noticed some breakages in test due to need to parse the HTML string first. Update the changelog to explicitly reflect this.


philss:
@kianmeng thank you!

kianmeng:
:partying_face: :partying_face: :partying_face: :partying_face: :partying_face:

PR #672, opened 2026-04-10
Fix typespecs: remove binary from Floki.attribute/3 typespec

preciz:
Binary as first arg no longer supported.

### DIFF
--- a/lib/floki.ex
+++ b/lib/floki.ex
@@ -614,7 +614,7 @@ defmodule Floki do
       ["e.corp"]
   """
 
-  @spec attribute(binary | html_tree | html_node, binary, binary) :: list
+  @spec attribute(html_tree | html_node, binary, binary) :: list
 
   def attribute(html, selector, attribute_name) do
     html


lib/floki.ex
```diff
@@ -614,7 +614,7 @@ defmodule Floki do
       ["e.corp"]
   """
 
-  @spec attribute(binary | html_tree | html_node, binary, binary) :: list
+  @spec attribute(html_tree | html_node, binary, binary) :: list
```
philss, line 617: This looks correct, but there is a problem: `html_node` can be an `html_text` too, which is a `String.t()`.

I'm not sure if we should bother about this, but we could have a `non_text_html_node` that better represents the type here. WDYT?

**EDIT:** we could also have a guard to prevent binaries here.

preciz: Due to new Elixir type checking I think it's less important now so we can just leave it like this for now

philss: OK, agreed.

philss:
@preciz Once again, thank you very much for your contributions! <2

I'm going to cut a new release next week, probably on Tuesday.

PR #679, opened 2026-04-14
Refactor attribute_values to use for comprehension with pattern matching

preciz:
Replaces the Enum.reduce implementation with a for comprehension
combined with a custom recursive get_attribute_value/2 helper.
This yields a 1.8x performance speed-up and is less lines of code.

### DIFF
--- a/lib/floki.ex
+++ b/lib/floki.ex
@@ -641,36 +641,19 @@ defmodule Floki do
   end
 
   defp attribute_values(elements, attr_name) do
-    values =
-      Enum.reduce(
-        elements,
-        [],
-        fn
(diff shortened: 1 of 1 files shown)


lib/floki.ex
```diff
@@ -641,37 +641,20 @@ defmodule Floki do
...
-          _, acc ->
-            acc
-        end
-      )
-
-    Enum.reverse(values)
+    for {_, attributes, _} <- elements,
+        value = get_attribute_value(attributes, attr_name),
+        not is_nil(value),
+        do: value
```
philss, line 647: Sweet! ✨

PR #685, opened 2026-04-17
Optimize HTMLTree updates and deletions

preciz:
- Replace macro `put_in/2` with explicit `Map.put/3` update syntax in `patch_nodes/2` for faster static map updates.
- Replace `Map.take/2` and `Map.values/1` pipeline with a list comprehension in `do_delete/3` for faster bulk node retrieval without intermediate allocations.

These optimizations provide an 5%-9% overall execution speed up to Floki's core mutation operations like Floki.find_and_update/3, Floki.filter_out/2, and Floki.attr/4 when manipulating large documents.

### DIFF
--- a/lib/floki/html_tree.ex
+++ b/lib/floki/html_tree.ex
@@ -140,10 +140,7 @@ defmodule Floki.HTMLTree do
   end
 
   defp do_delete(tree, [], stack_ids) do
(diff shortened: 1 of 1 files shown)


preciz:
Will open another PR instead.

PR #691, opened 2026-04-22
Optimize selector traversal

preciz:
Avoid extra reversals while preserving descendant order.


preciz:
@philss I plan no further changes so it could be a good time to make a release if you think alike.

